PR #357, opened 2024-10-04
♻️ (core) [DSDK-503]: Improve WebHID reconnection logic & internal logic

ofreyssinet-ledger:
#### Overall improvements:
- `sendApdu`: it will now wait for a reconnection (or timeout) in case the connection has been lost, before trying to call `sendReport` on the `HIDDevice`.
- Multi device support: the previous implementation was not able to handle multiple devices of the same model as we can't get a unique identifier from a device.
  - Previously, successively connecting 2 devices of the same model would break the 1st of the 2 sessions, it's not the case anymore.
- Reconnection logic: it is now more robust.
- Overall I hope that the code is now easier to read and maintain.

#### Explanation of the changes:
Rework `WebUsbHidTransport` internal logic & reconnection logic:
- Stop relying on our own generated `deviceId` for internal mapping of discovered/connected devices, rely on the identity of `HIDDevice` instead:
   - we only have a `Map<DeviceId,WebHidInternalDevice>` (each `WebHidInternalDevice` contains a `InternalDiscoveredDevice` `HIDDevice`) to handle `DeviceId` (which is needed by `connect` `disconnect`) .
   - A `UsbHidDeviceConnection` can be stored in ONE of two places:
  ```ts
    /** Maps all *connected* HIDDevice to their UsbHidDeviceConnection */
  private _deviceConnectionsByHidDevice: Map<
    HIDDevice,
    UsbHidDeviceConnection
  > = new Map();
  /** Set of all the UsbHidDeviceConnection for which HIDDevice has been disconnected, so they are waiting for a reconnection */
  private _deviceConnectionsPendingReconnection: Set<UsbHidDeviceConnection> =
    new Set();
  ```
-  there are mainly 3 scenarios for connection/disconnection:
   - initiating a connection `connect({deviceId})`:
     - we create the `UsbHidDeviceConnection` and store it in  `_deviceConnectionsByHidDevice`
       - this is possible because `HIDDevice` identity is stable as long as the `HIDDevice` is connected, the navigator.hid API will not return a new `HIDDevice` for the same physical device.
   - disconnecting `disconnect({deviceId})`:
     - we find the internal device, which gives us the `HIDDevice`, and we find the associated `UsbHidDeviceConnection` in `_deviceConnectionsByHidDevice`.
   - detection of a disconnection of an `HIDDevice` (can be due to unplugging the device OR app switching):
     - we remove the `HIDDevice` from `_deviceConnectionsByHidDevice` and push the associated `UsbHidDeviceConnection` to `_deviceConnectionsPendingReconnection`.
     - the `UsbHidDeviceConnection` is notified about the disconnection (`.lostConnection()`) and internally starts a timer.
     - IF the timer runs out, the connection is terminated and the `UsbHidDeviceConnection` is removed from `_deviceConnectionsPendingReconnection` and everything is cleaned up
     - ELSE IF a similar device model gets connected, we kill the timer and move back the `UsbHidDeviceConnection` in `_deviceConnectionsByHidDevice`.
- Disconnection handlers & timers have been moved from `WebHidTransport` to `UsbHidDeviceConnection` because they are connection-specific
  - easier to read & maintain.
- `UsbHidDeviceConnection.sendApdu()` will now wait for a reconnection (or timeout) before calling `HIDDevice.sendReport()` in case the connection has been lost. This fix is needed for the integration of the LDMK into Ledger Live Desktop, to properly handle firmware update flows.

### ❓ Context

- **JIRA or GitHub link**: [DSDK-503] 

### ✅ Checklist

Pull Requests must pass the CI and be code reviewed. Set as Draft if the PR is not ready.

- [x] **Covered by automatic tests.** 
  - device reconnection to test in the sample app:
      - connection doesn't get lost when opening/closing apps, for instance test the open app device action from one app to another.
      - actual disconnection of a device is handled properly (disconnection after 2s timeout)
---

[DSDK-503]: https://ledgerhq.atlassian.net/browse/DSDK-503?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/core/src/internal/usb/data/UsbHidConfig.ts
+++ b/packages/core/src/internal/usb/data/UsbHidConfig.ts
@@ -1,4 +1,4 @@
 // [SHOULD] Move it to device-model module
 export const LEDGER_VENDOR_ID = 0x2c97;
 export const FRAME_SIZE = 64;
-export const RECONNECT_DEVICE_TIMEOUT = 2000;
+export const RECONNECT_DEVICE_TIMEOUT = 6000; // in some cases, when opening/closing an app, it takes up to 6s between the HID "disconnect" and "connect" events
--- a/packages/core/src/internal/usb/transport/UsbHidDeviceConnection.ts
+++ b/packages/core/src/internal/usb/transport/UsbHidDeviceConnection.ts
@@ -1,59 +1,82 @@
 import { inject } from "inversify";
-import { Either, Left, Maybe, Right } from "purify-ts";
+import { Either, Left, Right } from "purify-ts";
 import { Subject } from "rxjs";
 
 import { CommandUtils } from "@api/command/utils/CommandUtils";
 import { ApduResponse } from "@api/device-session/ApduResponse";
 import { SdkError } from "@api/Error";
+import { DeviceId } from "@api/types";
 import { ApduReceiverService } from "@internal/device-session/service/ApduReceiverService";
 import { ApduSenderService } from "@internal/device-session/service/ApduSenderService";
 import { loggerTypes } from "@internal/logger-publisher/di/loggerTypes";
 import type { LoggerPublisherService } from "@internal/logger-publisher/service/LoggerPublisherService";
-import { ReconnectionFailedError } from "@internal/usb/model/Errors";
+import { RECONNECT_DEVICE_TIMEOUT } from "@internal/usb/data/UsbHidConfig";
+import {
+  HidSendReportError,
+  ReconnectionFailedError,
+} from "@internal/usb/model/Errors";
 
 import { DeviceConnection } from "./DeviceConnection";
 
 type UsbHidDeviceConnectionConstructorArgs = {
   device: HIDDevice;
+  deviceId: DeviceId;
   apduSender: ApduSenderService;
   apduReceiver: ApduReceiverService;
+  onConnectionTerminated: () => void;
 };
 
+/**
+ * Class to manage the connection with a USB HID device.
+ * It sends APDU commands to the device and receives the responses.
+ * It handles temporary disconnections and reconnections.
+ */
 export class UsbHidDeviceConnection implements DeviceConnection {
   private _device: HIDDevice;
+  private _deviceId: DeviceId;
   private readonly _apduSender: ApduSenderService;
   private readonly _apduReceiver: ApduReceiverService;
-  private _sendApduSubject: Subject<ApduResponse>;
+  private _sendApduSubject: Subject<ApduResponse> = new Subject();
   private readonly _logger: LoggerPublisherService;
-  private _settleReconnectionPromise: Maybe<{
-    resolve(): void;
-    reject(err: SdkError): void;
-  }> = Maybe.zero();
+
+  /** Callback to notify the connection termination */
+  private _onConnectionTerminated: () => void;
+  /** Subject to notify the reconnection status */
+  private reconnectionSubject: Subject<"success" | SdkError> = new Subject();
+  /** Flag to indicate if the connection is waiting for a reconnection */
+  private waitingForReconnection = false;
+  /** Timeout to wait for the device to reconnect */
+  private lostConnectionTimeout: NodeJS.Timeout | null = null;
+  /** Flag to indicate if the connection is terminated */
+  private terminated = false;
 
   constructor(
-    { device, apduSender, apduReceiver }: UsbHidDeviceConnectionConstructorArgs,
+    {
+      device,
+      deviceId,
+      apduSender,
+      apduReceiver,
+      onConnectionTerminated,
+    }: UsbHidDeviceConnectionConstructorArgs,
     @inject(loggerTypes.LoggerPublisherServiceFactory)
     loggerServiceFactory: (tag: string) => LoggerPublisherService,
   ) {
     this._apduSender = apduSender;
     this._apduReceiver = apduReceiver;
-    this._sendApduSubject = new Subject();
+    this._onConnectionTerminated = onConnectionTerminated;
     this._logger = loggerServiceFactory("UsbHidDeviceConnection");
     this._device = device;
     this._device.oninputreport = (event) => this.receiveHidInputReport(event);
+    this._deviceId = deviceId;
+    this._logger.info("🔌 Connected to device");
   }
 
   public get device() {
     return this._device;
   }
 
-  public set device(device: HIDDevice) {
-    this._device = device;
-    this._device.oninputreport = (event) => this.receiveHidInputReport(event);
-
-    this._settleReconnectionPromise.ifJust(() => {
-      this.reconnected();
-    });
+  public get deviceId() {
+    return this._deviceId;
   }
 
   async sendApdu(
@@ -72,7 +95,8 @@ export class UsbHidDeviceConnection implements DeviceConnection {
         this._sendApduSubject.subscribe({
           next: async (r) => {
             if (triggersDisconnection && CommandUtils.isSuccessResponse(r)) {
-              const reconnectionRes = await this.setupWaitForReconnection();
+              // Anticipate the disconnection and wait for the reconnection before resolving
+              const reconnectionRes = await this.waitForReconnection();
               reconnectionRes.caseOf({
                 Left: (err) => resolve(Left(err)),
                 Right: () => resolve(Right(r)),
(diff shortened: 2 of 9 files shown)


packages/core/src/internal/usb/transport/WebUsbHidTransport.ts
```diff
@@ -191,9 +188,23 @@ export class WebUsbHidTransport implements UsbHidTransport {
             throw error;
           },
           Right: (hidDevices) => {
-            this._logger.info(`Got access to ${hidDevices.length} HID devices`);
+            this._logger.info(
+              `Got access too ${hidDevices.length} HID devices`,
```
aussedatlo, line 192: [COULD] typo `too`

aussedatlo (approved):
Nice work 🔥 
We have a lot of syntax inconsistency between using `HID` and `Hid`. I suggest we stick to using `Hid` in our code, and only use `HID` when we have no other option.

packages/core/src/internal/usb/data/UsbHidConfig.ts
```diff
@@ -1,4 +1,4 @@
 // [SHOULD] Move it to device-model module
 export const LEDGER_VENDOR_ID = 0x2c97;
 export const FRAME_SIZE = 64;
-export const RECONNECT_DEVICE_TIMEOUT = 2000;
+export const RECONNECT_DEVICE_TIMEOUT = 6000; // in some cases, when opening/closing an app, it takes up to 6s between the HID "disconnect" and "connect" events
```
valpinkman, line 4: [ASK] Is this on all devices ?

aussedatlo (approved):

packages/core/src/internal/usb/transport/UsbHidDeviceConnection.ts
```diff
@@ -1,59 +1,82 @@
...
   private readonly _logger: LoggerPublisherService;
-  private _settleReconnectionPromise: Maybe<{
-    resolve(): void;
-    reject(err: SdkError): void;
-  }> = Maybe.zero();
+
+  /** Callback to notify the connection termination */
+  private _onConnectionTerminated: () => void;
+  /** Subject to notify the reconnection status */
+  private reconnectionSubject: Subject<"success" | SdkError> = new Subject();
```
valpinkman, line 45: [ASK] Do we want a specific type for `"success"` here ?

packages/core/src/internal/usb/transport/UsbHidDeviceConnection.ts
```diff
@@ -126,26 +158,55 @@ export class UsbHidDeviceConnection implements DeviceConnection {
...
-        resolve: () => resolve(Right(undefined)),
-        reject: (error: SdkError) => resolve(Left(error)),
+      const sub = this.reconnectionSubject.subscribe({
+        next: (res) => {
+          if (res === "success") {
+            resolve(Right(undefined));
+          } else {
+            resolve(Left(res));
+          }
+          sub.unsubscribe();
```
valpinkman, line 172: [ASK] Is this to make sure we only get one device at a time ? (the unsubscribe)

packages/core/src/internal/usb/transport/UsbHidDeviceConnection.ts
```diff
@@ -126,26 +158,55 @@ export class UsbHidDeviceConnection implements DeviceConnection {
...
   public disconnect() {
-    this._settleReconnectionPromise.ifJust((promise) => {
-      promise.reject(new ReconnectionFailedError());
-      this._settleReconnectionPromise = Maybe.zero();
-    });
+    this._logger.info("🔚 Disconnect");
+    if (this.lostConnectionTimeout) clearTimeout(this.lostConnectionTimeout);
+    this.terminated = true;
+    this._onConnectionTerminated();
+    this.reconnectionSubject.next(new ReconnectionFailedError());
```
valpinkman, line 210: [ASK] I'm not sure when disconnect is called, but it seems that we always bubble up a reconnection error on disconnect. 
What if I just unplugged my device ? Maybe we don't have a way to determine that, in this case I can understand that we always go with a Left of ReconnectionFailedError.

packages/core/src/internal/usb/transport/WebUsbHidTransport.ts
```diff
@@ -60,14 +62,13 @@ export class WebUsbHidTransport implements UsbHidTransport {
...
   ) {
-    this._internalDevicesById = new Map();
-    this._disconnectionHandlersByHidId = new Map();
-    this._deviceConnectionByHidId = new Map();
     this._connectionListenersAbortController = new AbortController();
     this._logger = loggerServiceFactory("WebUsbHidTransport");
     this._usbHidDeviceConnectionFactory = usbHidDeviceConnectionFactory;
 
     this.hidApi.map((hidApi) => {
+      // FIXME: we should not override the global navigator.hid.onconnect and navigator.hid.ondisconnect but instead use addEventListener
```
valpinkman, line 70: Strongly agreeing with this comment

packages/core/src/internal/usb/transport/WebUsbHidTransport.ts
```diff
@@ -350,8 +362,26 @@ export class WebUsbHidTransport implements UsbHidTransport {
...
+    this._logger.debug("_internalDisconnect", {
+      data: { connectedDevice: params },
+    });
+    const { internalDevice } = params;
+
+    try {
+      this._internalDevicesById.delete(internalDevice.id);
+      this._deviceConnectionsByHidDevice.delete(internalDevice.hidDevice);
+      await internalDevice.hidDevice.close();
+      return Right(void 0);
```
valpinkman, line 377: Still not a fan of returning the `void 0` but not problematic either

valpinkman (approved):
I have left a few comments / questions, but overall it's good to go

packages/core/src/internal/usb/data/UsbHidConfig.ts
ofreyssinet-ledger: Yes for now, we can probably fine tune it later 👍

packages/core/src/internal/usb/transport/UsbHidDeviceConnection.ts
ofreyssinet-ledger: meh, it's just for some internal notifications 🤷‍♂️ i'd say KISS

packages/core/src/internal/usb/transport/UsbHidDeviceConnection.ts
ofreyssinet-ledger: It's just to cleanup the subscription once the promise is resolved.
For "one device at a time" anyway we have this relation: 1 `UsbHidDeviceConnection` <-> 1 physical device

packages/core/src/internal/usb/transport/UsbHidDeviceConnection.ts
ofreyssinet-ledger: `disconnect` is called either externally (through calling `sdk.disconnect()`) or internally when the `HIDDevice` has been disconnected and the timeout is over.
The `reconnectionSubject` is only observed when an APDU is (trying to get) sent while a reconnection is pending. So only in that case will the error bubble up.

packages/core/src/internal/usb/transport/WebUsbHidTransport.ts
ofreyssinet-ledger: Fixed in my upcoming PR regarding listening to known devices 👍

packages/core/src/internal/usb/transport/WebUsbHidTransport.ts
ofreyssinet-ledger: Would returning `Right(undefined)` be better ? I'm doing this in my next PR

packages/core/src/internal/usb/transport/UsbHidDeviceConnection.ts
valpinkman: perfect

packages/core/src/internal/usb/transport/WebUsbHidTransport.ts
valpinkman: 🔥

packages/core/src/internal/usb/transport/WebUsbHidTransport.ts
jiyuzhuang: I think that `void 0` is the best way if we surely want to return a `undefined` value.

PR #392, opened 2024-10-11
✨ (core) [DSDK-504]: Listen to known devices (WebHID for now) & display them in sample app

ofreyssinet-ledger:
### 📝 Description

This new feature allows to observe an array of available devices, as `DiscoveredDevice`.

For WebHID, it means that for devices for which the user has already granted an authorization to the app (via `navigator.hid.requestDevice()`), plugging in a device will make it appear in that array, and unplugging it will make it disappear.

In case of reconnection (in an open/close app for instance), internally, the `HIDDevice` will be a different instance, but we keep the same `DeviceId` so that from the app's point of view, we can identify which of the `DiscoveredDevice` already has an ongoing session.

https://github.com/user-attachments/assets/ce5c5598-f2ff-4183-9177-466644ab7cee

### ❓ Context

- **JIRA or GitHub link**: [[DSDK-504]](https://ledgerhq.atlassian.net/jira/software/c/projects/DSDK/boards/817?assignee=6144f82ee7c3280070b2507c&selectedIssue=DSDK-504)

- **Feature**:

### ✅ Checklist

  - (core) new use case listen to known devices
  - (sample) display of the known devices in a new section of the side bar, with a connect button (which is grayed out when the device is already connected).
  - (core) small refacto of the internal logic of `WebUsbHidTransport` for discovered devices

---

### DIFF
--- a/apps/sample/src/providers/DeviceSdkProvider/index.tsx
+++ b/apps/sample/src/providers/DeviceSdkProvider/index.tsx
@@ -9,62 +9,90 @@ import {
 } from "@ledgerhq/device-management-kit";
 import { FlipperSdkLogger } from "@ledgerhq/device-management-kit-flipper-plugin-client";
 
-import { usePrevious } from "@/hooks/usePrevious";
+import { useHasChanged } from "@/hooks/useHasChanged";
 import { useSdkConfigContext } from "@/providers/SdkConfig";
 
-const webLogsExporterLogger = new WebLogsExporterLogger();
+const SdkContext = createContext<DeviceSdk | null>(null);
+const LogsExporterContext = createContext<WebLogsExporterLogger | null>(null);
 
-const defaultSdk = new DeviceSdkBuilder()
-  .addLogger(new ConsoleLogger())
-  .addTransport(BuiltinTransports.BLE)
-  .addTransport(BuiltinTransports.USB)
-  .addLogger(webLogsExporterLogger)
-  .addLogger(new FlipperSdkLogger())
-  .build();
+function buildDefaultSdk(logsExporter: WebLogsExporterLogger) {
+  return new DeviceSdkBuilder()
+    .addTransport(BuiltinTransports.USB)
+    .addTransport(BuiltinTransports.BLE)
+    .addLogger(new ConsoleLogger())
+    .addLogger(logsExporter)
+    .addLogger(new FlipperSdkLogger())
+    .build();
+}
 
-const SdkContext = createContext<DeviceSdk>(defaultSdk);
+function buildMockSdk(url: string, logsExporter: WebLogsExporterLogger) {
+  return new DeviceSdkBuilder()
+    .addTransport(BuiltinTransports.MOCK_SERVER)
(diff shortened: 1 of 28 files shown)


jdabbech-ledger (changes requested):
Good rework of the transport hid
A lot of modifications in the sample app sdk provider as well, but there are few bugs when switching transport between mock and hid (by connecting device in the meantime). Moreover some e2e tests seem to be broken

packages/core/src/internal/transport/usb/transport/WebUsbHidTransport.ts
```diff
@@ -110,13 +108,111 @@ export class WebUsbHidTransport implements Transport {
...
+   * Listen to known devices (devices to which the user has granted access)
+   */
+  public listenToKnownDevices(): Observable<InternalDiscoveredDevice[]> {
+    this.updateInternalDiscoveredDevices();
+    return this._internalDiscoveredDevices.pipe(
+      map((devices) => devices.map(({ hidDevice, ...device }) => device)),
+    );
+  }
+
+  private async updateInternalDiscoveredDevices() {
```
jdabbech-ledger, line 192: [could] specify return value

packages/core/src/internal/transport/usb/transport/WebUsbHidTransport.ts
```diff
@@ -110,13 +108,111 @@ export class WebUsbHidTransport implements Transport {
...
+    await this.getDevices().then((either) => {
```
jdabbech-ledger, line 193: [ask] wouldn't it be clearer to not use `.then` if using `await` syntax?
[could] use EitherAsync

apps/sample/src/providers/DeviceSdkProvider/index.tsx
```diff
@@ -10,46 +10,80 @@ import {
...
+        sdk: buildDefaultSdk(logsExporter),
+        logsExporter,
+      }));
     }
   }, [mockServerEnabled, url]);
 
-  return <SdkContext.Provider value={sdk}>{children}</SdkContext.Provider>;
+  useEffect(() => {
+    return () => {
+      state.sdk.close();
```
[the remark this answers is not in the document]
ofreyssinet-ledger: The error when enabling mock server was due to something else. The part of the app listen to each session state was still calling getDeviceSessionState() with a invalid sessionId (when the SDK has been closed) and it was uncaught in a useEffect so it was actually crashing the app.
And there's probably some error boundary in Next remounting the app, but the state was not preserved, hence the fact that you had to toggle the mock server mode once again.
It's not happening anymore because now in the `DeviceSessionsProvider` we cleanup the state whenever the `sdk` obtained through `useSdk` changes.

The "it's called called too many times" is because of React Strict Mode in dev mode which runs effects an extra time. It's working fine now 👍

jdabbech-ledger (approved):

valpinkman (approved):

aussedatlo (approved):

PR #397, opened 2024-10-11
:bug: (context-module) [NO-ISSUE]: Fix CAL test signatures for EIP 712

paoun-ledger:
### 📝 Description

While testing full clear signing flow for the demo, we noticed two issues fixed in that PR:
* CAL don't necessarily return `prod` and `test` signatures. Prod CAL may return only prod signatures, and test CAL only the test signatures.
* When working with test signatures for EIP 712, we should also retrieve test signatures for the tokens. To do that, we should use new CAL api with already crafter descriptors

### ❓ Context

- **JIRA or GitHub link**:

- **Feature**:

### ✅ Checklist

  - _list of the changes_

---


packages/signer/context-module/src/token/data/TokenDto.ts
```diff
@@ -1,5 +1,14 @@
+export type TokenSignatures = {
+  test?: string;
```
aussedatlo, line 2: [ASK] why `test` and `prod` are optional here ? we can have a `TokenDescriptor` without signature ?

packages/signer/context-module/src/typed-data/data/HttpTypedDataDataSource.ts
```diff
@@ -89,29 +89,29 @@ export class HttpTypedDataDataSource implements TypedDataDataSource {
         if (this.isInstructionContractInfo(field)) {
           messageInfo = {
             displayName: field.display_name,
-            signature: field.signatures[this.config.cal.mode],
+            signature: field.signatures[this.config.cal.mode]!,
```
aussedatlo, line 92: [SHOULD] avoid using `!` and add a check to see if the signature is available, or return an error

aussedatlo: Ok this is already verified by the typeguard, but if you need it it's probably because the type `Signature` is not efficient enough

packages/signer/context-module/src/token/data/TokenDto.ts
paoun-ledger: we can have only test, or only prod

packages/signer/context-module/src/typed-data/data/HttpTypedDataDataSource.ts
paoun-ledger: done

aussedatlo (approved):

PR #423, opened 2024-10-17
🚧 (keyring-eth) [DSDK-525]: Add ProvideTransactionGenericContext Task

aussedatlo:
### 📝 Description

Add `ProvideTransactionGenericContextTask`

### ❓ Context

- **JIRA or GitHub link**: [DSDK-525]

- **Feature**:

### ✅ Checklist

  - _list of the changes_

---

[DSDK-525]: https://ledgerhq.atlassian.net/browse/DSDK-525?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/signer/keyring-eth/src/internal/app-binder/task/ProvideTransactionGenericContextTask.ts
+++ b/packages/signer/keyring-eth/src/internal/app-binder/task/ProvideTransactionGenericContextTask.ts
@@ -5,9 +5,10 @@ import {
 import {
   type CommandErrorResult,
   CommandResult,
+  CommandResultFactory,
   type InternalApi,
+  InvalidStatusWordError,
   isSuccessCommandResult,
-  type SdkError,
 } from "@ledgerhq/device-management-kit";
 import { Just, Maybe, Nothing } from "purify-ts";
 
@@ -30,7 +31,10 @@ import {
 import { StoreTransactionCommand } from "@internal/app-binder/command/StoreTransactionCommand";
 import { PayloadUtils } from "@internal/shared/utils/PayloadUtils";
 
-import { SendCommandInChunksTask } from "./SendCommandInChunksTask";
+import {
+  SendCommandInChunksTask,
+  SendCommandInChunksTaskArgs,
+} from "./SendCommandInChunksTask";
 
 export type ProvideTransactionGenericContextTaskArgs = {
   serializedTransaction: Uint8Array;
@@ -38,20 +42,6 @@ export type ProvideTransactionGenericContextTaskArgs = {
   transactionFieldDescription: Record<string, string>;
   metadatas: Record<string, ClearSignContextSuccess>;
 };
-
-/**
- * Temporary error type to be used in the `ProvideTransactionGenericContextTask` in order to not forget to handle the error cases.
- */
-export class ProvideTransactionGenericContextTaskError implements SdkError {
-  readonly _tag = "ProvideTransactionGenericContextTaskError";
-  readonly originalError: Error;
-
-  constructor(message?: string) {
-    this.originalError = new Error(
-      message ?? "Unknow error in ProvideTransactionGenericContextTaskError",
-    );
-  }
-}
 
 export type ProvideTransactionGenericContextTaskErrorCodes =
   | void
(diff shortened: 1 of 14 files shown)


packages/signer/keyring-eth/src/internal/app-binder/task/SendCommandInChunksTask.ts
```diff
@@ -0,0 +1,60 @@
...
+type SendCommandInChunksTaskArgs<T> = {
+  data: Uint8Array;
+  commandFactory: <V extends ChunkableCommandArgs & Record<string, unknown>>(
+    args: ChunkableCommandArgs,
+  ) => Command<T, V>;
+};
+
+type ChunkableCommandArgs = {
+  chunkedData: Uint8Array;
+  isFirstChunk: boolean;
```
valpinkman, line 21: [NOTE] It seems that we are not using `isFirstChunk` here in the end as we get it from `i === 0` on line 42.

aussedatlo: It's a bit tricky, this line is necessary:
the factory constructs the command using the arguments provided by the task. This type represents the structure of those arguments passed from the task to the command.

packages/signer/keyring-eth/src/internal/app-binder/task/SendSignPersonalMessageTask.ts
```diff
@@ -49,25 +49,24 @@ export class SendSignPersonalMessageTask {
...
 
-      if (!isSuccessCommandResult(result)) {
-        return result;
-      }
+    if (!isSuccessCommandResult(result)) {
+      return result;
     }
 
-    if (isSuccessCommandResult(result) && result.data.isJust()) {
+    if (result.data.isJust()) {
```
jiyuzhuang, line 69: Can you please test this snippet, I think that we should take the advantage of making chaining methods. 
```suggestion
    return result.data
      .ifJust(data => CommandResultFactory({ data }))
      .ifNothing(() => CommandResultFactory({
        error: new InvalidStatusWordError("no signature returned"),	
      }));
```

jiyuzhuang:
It looks like good for me, globally:
1. I think we should make chaining methods when using of purify functions, I've added a comment in one place, but should check all the similar cases and chain the methods if possible;
2. Can you please add some comments in the utility methods ? This will make it more understandable. 😄

packages/signer/keyring-eth/src/internal/app-binder/task/SendSignPersonalMessageTask.ts
aussedatlo: This chaining code is not working, because we need to return the commandFactory,

but we can use a `mapOrDefault`operator instead:

```ts
return result.data.mapOrDefault(
      (data) => CommandResultFactory({ data }),
      CommandResultFactory({
        error: new InvalidStatusWordError("no signature returned"),
      }),
    );
```

packages/signer/keyring-eth/src/internal/app-binder/task/ProvideTransactionGenericContextTask.ts
```diff
@@ -0,0 +1,209 @@
...
+        }).run();
+      }
+      case ClearSignContextType.EXTERNAL_PLUGIN: {
+        throw new ProvideTransactionGenericContextTaskError(
+          "The context type [EXTERNAL_PLUGIN] is not valid here",
+        );
+      }
+      default: {
+        const uncoveredType: never = type;
+        throw new ProvideTransactionGenericContextTaskError(
```
jdabbech-ledger, line 203: [could] use CommandErrorResult instead of throwing error

packages/signer/keyring-eth/src/internal/app-binder/task/SendCommandInChunksTask.ts
```diff
@@ -0,0 +1,60 @@
...
+
+      // return the last result
+      if (isLastChunk) {
+        return CommandResultFactory({
+          data: result.data,
+        });
+      }
+    }
+
+    throw new InvalidStatusWordError("No result");
```
jdabbech-ledger, line 58: [could] use CommandErrorResult

packages/signer/keyring-eth/src/internal/shared/utils/PayloadUtils.ts
```diff
@@ -0,0 +1,22 @@
...
+} from "@ledgerhq/device-management-kit";
+
+export class PayloadUtils {
+  static PAYLOAD_LENGTH_BYTES = 2;
+
+  static getBufferFromPayload(payload: string): Uint8Array {
+    const buffer = hexaStringToBuffer(payload);
+
+    if (buffer === null || buffer.length === 0) {
+      throw new HexaStringEncodeError("Invalid payload");
```
jdabbech-ledger, line 14: [could] use Either

packages/signer/keyring-eth/src/internal/app-binder/task/SendSignPersonalMessageTask.ts
jiyuzhuang: Yes, you're right. The `ifXXXX` method always return `this`. But let's make a chain. 😄

packages/signer/keyring-eth/src/internal/app-binder/task/ProvideTransactionGenericContextTask.ts
aussedatlo: Yes !

packages/signer/keyring-eth/src/internal/shared/utils/PayloadUtils.ts
aussedatlo: Not sure about this one, keeping a util without purify dependency could be good as well, but it's a good point to avoid using throw here.

What do you think about changing the return type to `Uint8Array | null` like the return type of `hexaStringToBuffer` ?

jdabbech-ledger: As it's an internal code it should be preferable to use Maybe or Either imo. 
But it's not a big deal obviously it works well and it would be simpler to externalise if needed without those dependencies.

jdabbech-ledger (approved):

PR #443, opened 2024-10-25
🚧 (core) [DSDK-560]: POC instantiate transport in sample app and setup di

jdabbech-ledger:
### 📝 Description

POC to instanciate transports from sample app and setup DI for logger and devicemodeldatasource.
- As the Transport is instanciated before the SDK DI container, we can't use logger in transport constructor. To fix this we could instantiate the transport in DefaultTransportService 
- Transports are added in Sdk constructor in order to have a possible init error thrown on `builder.build`
- Commit https://github.com/LedgerHQ/device-sdk-ts/commit/eb585a75d54b636347751e2d36aa964761711f11 is just to have core builded and testable in sample app. Will be deleted and tests fixed if this poc is ok

To review this poc, please focus on those commits: https://github.com/LedgerHQ/device-sdk-ts/pull/443/commits/cb414b29dab984076c1305dd3ac8d3975defa612 https://github.com/LedgerHQ/device-sdk-ts/pull/443/commits/01e761ab34842121bc36451b02dcec2aec83b079 https://github.com/LedgerHQ/device-sdk-ts/pull/443/commits/51a05139c5714fa963b2bf6d7503f33ac464bd0f

### ❓ Context

- **JIRA or GitHub link**: [DSDK-560]

- **Feature**:

### ✅ Checklist

  - _list of the changes_

---


smartine-ledger (commented):
Ok for the POC but personally I am not a fan of this implementation which complicate the code (`setter` / `getter` in the `Transport` interface + `TransportService` which should be renamed to `TransportFactory` 🙂 + errors cases to handle (same transport added twice...)
As I said at the beginning and  I do not understand why do we let the user inject the transport instead of enabling / disabling them (which it would avoid a lot complication IMHO) 🤷‍♂️

jdabbech-ledger:
Thanks for your review, we don't enable/disable transport inside the sdk because we don't want to build a sdk that includes react-native transport when we use it in a web platform for example.
We then have several choices to inject them from an app:
- by instance (which is done in this PR and requires transport setters as logger and framers are created after transport)
- by signature (no setter required)
- by builder (no setter required)

smartine-ledger:
> Thanks for your review, we don't enable/disable transport inside the sdk because we don't want to build a sdk that includes react-native transport when we use it in a web platform for example. We then have several choices to inject them from an app:
> 
> * by instance (which is done in this PR and requires transport setters as logger and framers are created after transport)
> * by signature (no setter required)
> * by builder (no setter required)

I don't get your point. Enable / Disable a transport does not mean including react-native transport. This just mean the user enable something and the DMK is responsible of getting / injecting everything the transport needs.

jdabbech-ledger:
By sdk I mean `@ledgerhq/device-management-kit` core package, my point is that the transport should be instantiated outside of this package. 
I just added a commit with builder for transport classes and without setters in transport interface.

smartine-ledger:
> By sdk I mean `@ledgerhq/device-management-kit` core package, my point is that the transport should be instantiated outside of this package. I just added a commit with builder for transport classes and without setters in transport interface.

The TransportBuilder does not change the problem IMHO. You're just adding "another layer"... Let's talk about it 😉

PR #475, opened 2024-11-05
Add hex messages as such when signing

iamacook:
### 📝 Description

Messages-to-sign are potentially hex. However, they are always passed as ASCII.

This adds an appropriate check to determine whether a message is hex, passing it as such if it is. The length of the message is also determined in a similar way.

### ❓ Context

- **JIRA or GitHub link**:

- **Feature**:

### ✅ Checklist

  - _list of the changes_

---

### DIFF
--- a/packages/signer/signer-eth/src/internal/app-binder/task/SendSignPersonalMessageTask.ts
+++ b/packages/signer/signer-eth/src/internal/app-binder/task/SendSignPersonalMessageTask.ts
@@ -2,8 +2,10 @@ import {
   ByteArrayBuilder,
   type CommandResult,
   CommandResultFactory,
+  hexaStringToBuffer,
   type InternalApi,
   InvalidStatusWordError,
+  isHexaString,
   isSuccessCommandResult,
 } from "@ledgerhq/device-management-kit";
 import { DerivationPathUtils } from "@ledgerhq/signer-utils";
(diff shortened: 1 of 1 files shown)


packages/signer/signer-eth/src/internal/app-binder/task/SendSignPersonalMessageTask.ts
```diff
@@ -43,9 +49,13 @@ export class SendSignPersonalMessageTask {
       builder.add32BitUIntToData(path);
     });
     // add message length
-    builder.add32BitUIntToData(message.length);
+    builder.add32BitUIntToData(messageLength);
     // add the message
-    builder.addAsciiStringToData(message);
+    if (isHexaString(message)) {
```
paoun-ledger, line 54: not sure it's the best solution because one may want to pass a string which look like a hex string, but woudn't want it to be decoded as a buffer...
Maybe a better solution would be to change the parameter of `signMessage` api to be `string | Uint8Array` so you can pass a byte array instead of utf8 string in your case. Would that be ok for you ?

paoun-ledger: here is an api proposal: https://github.com/LedgerHQ/device-sdk-ts/pull/476/files

iamacook: Good point. I agree that expanding the type would make the API more flexible and prevent unintended decoding. Your proposed solution looks good to me. I’ll go ahead and close this.

iamacook:
Superseeded by https://github.com/LedgerHQ/device-sdk-ts/pull/476.

PR #508, opened 2024-11-19
(signer-eth) [DSDK-582]: Ethereum/Bitcoin internal transaction mappers - Improve typeguards to not rely on instanceof

jdabbech-ledger:
### 📝 Description

Perform a check on the properties of the transaction objects to determine if they are valid.

### ❓ Context

- **JIRA or GitHub link**: [DSDK-582]

- **Feature**:

### ✅ Checklist

  - _list of the changes_

---


valpinkman (approved):
nice

packages/signer/signer-eth/src/internal/transaction/service/mapper/EthersV6TransactionMapper.ts
```diff
@@ -29,6 +33,21 @@ export class EthersV6TransactionMapper implements TransactionMapper {
...
+      (transaction.gasPrice === null || isBigNumberish(transaction.gasPrice)) &&
+      (transaction.value === null || isBigNumberish(transaction.value)) &&
+      (transaction.chainId === null ||
+        isBigNumberish(typeof transaction.chainId)) &&
+      (transaction.type === null || typeof transaction.type === "number")
+    );
   }
 }
+
+const isBigNumberish = (number: unknown): number is BigNumberish =>
```
jiyuzhuang, line 50: There is an [exported function](https://github.com/ethers-io/ethers.js/blob/f97b92bbb1bde22fcc44100af78d7f31602863ab/packages/bignumber/src.ts/bignumber.ts#L27) in the lib, maybe take advantage of it.

jdabbech-ledger: Oh good catch thanks

jdabbech-ledger: Actually it needs to add the dependancy of `@ethersproject/bignumber` and this typeguard is then exported from `@ethersproject/bignumber/lib/bignumber`. 
As the transaction is already created with v5 or v6 type, is it really necessary to add this check here ? @aussedatlo @paoun-ledger  
```ts
 return (value != null) && (
        BigNumber.isBigNumber(value) ||
        (typeof(value) === "number" && (value % 1) === 0) ||
        (typeof(value) === "string" && !!value.match(/^-?[0-9]+$/)) ||
        isHexString(value) ||
        (typeof(value) === "bigint") ||
        isBytes(value)
);
```

jiyuzhuang: @aussedatlo I'm not sure whether some warnings will appear in the codebase when removing the type guard, then if the property `chainId` is present with wrong value type (modified by lib users for example), there will be type error, so I think it's better to keep the type guard. About the import, I didn't know that it's not exported in `ethers-v*`, if you think adding the dependency for just one import is a bit overkill, I think original code is also OK in this case.

PR #512, opened 2024-11-20
✨ (btc-signer) [DSDK-465]: Add GetWalletAddressCommand

fAnselmi-Ledger:
### 📝 Description

Added GetWalletAddressCommand

### ❓ Context

- **JIRA or GitHub link**: [DSDK-465](https://ledgerhq.atlassian.net/browse/DSDK-465)

### ✅ Checklist

  - _list of the changes_

---

[DSDK-465]: https://ledgerhq.atlassian.net/browse/DSDK-465?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.changeset/quiet-feet-leave.md
+++ b/.changeset/quiet-feet-leave.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/device-signer-kit-bitcoin": minor
+---
+
+Add GetWalletAddressCommand
--- a/packages/signer/signer-btc/src/internal/app-binder/command/GetWalletAddressCommand.test.ts
+++ b/packages/signer/signer-btc/src/internal/app-binder/command/GetWalletAddressCommand.test.ts
@@ -1,0 +1,149 @@
+import {
+  ApduResponse,
+  CommandResultFactory,
+  isSuccessCommandResult,
+} from "@ledgerhq/device-management-kit";
+
+import {
+  BitcoinAppCommandError,
+  bitcoinAppErrors,
+} from "./utils/bitcoinAppErrors";
+import {
+  GetWalletAddressCommand,
+  type GetWalletAddressCommandArgs,
+} from "./GetWalletAddressCommand";
+
+describe("GetWalletAddressCommand", () => {
+  let command: GetWalletAddressCommand;
+  const defaultArgs: GetWalletAddressCommandArgs = {
+    display: true,
+    walletId: Uint8Array.from("walletIdBuffer", (c) => c.charCodeAt(0)),
+    walletHmac: Uint8Array.from("walletHmacBuffer", (c) => c.charCodeAt(0)),
+    change: false,
+    addressIndex: 0x00000000,
+  };
+
+  beforeEach(() => {
+    command = new GetWalletAddressCommand(defaultArgs);
+    jest.clearAllMocks();
+    jest.requireActual("@ledgerhq/device-management-kit");
+  });
+
+  describe("getApdu", () => {
+    it("should return correct APDU for default arguments", () => {
+      const apdu = command.getApdu();
+      const expectedApdu = Uint8Array.from([
+        0xe1, // CLA
+        0x03, // INS
+        0x00, // P1
+        0x01, // P2
+        0x24, // Length of data: 36 bytes
+        0x01, // display: true
+        ...Uint8Array.from("walletIdBuffer", (c) => c.charCodeAt(0)),
+        ...Uint8Array.from("walletHmacBuffer", (c) => c.charCodeAt(0)),
+        0x00, // change: false
+        0x00,
+        0x00,
+        0x00,
+        0x00, // addressIndex: 0x00000000
+      ]);
+      expect(apdu.getRawApdu()).toEqual(expectedApdu);
+    });
+
+    it("should return correct APDU for different arguments", () => {
+      const args: GetWalletAddressCommandArgs = {
+        display: false,
+        walletId: Uint8Array.from("anotherWalletId", (c) => c.charCodeAt(0)),
+        walletHmac: Uint8Array.from("anotherWalletHmac", (c) =>
+          c.charCodeAt(0),
+        ),
+        change: true,
+        addressIndex: 0x00000005,
+      };
+      command = new GetWalletAddressCommand(args);
+      const apdu = command.getApdu();
+      const expectedApdu = Uint8Array.from([
+        0xe1, // CLA
+        0x03, // INS
+        0x00, // P1
+        0x01, // P2
+        0x26, // Length of data
+        0x00, // display: false
+        ...Uint8Array.from("anotherWalletId", (c) => c.charCodeAt(0)),
+        ...Uint8Array.from("anotherWalletHmac", (c) => c.charCodeAt(0)),
+        0x01, // change: true
+        0x00,
+        0x00,
+        0x00,
+        0x05, // addressIndex: 0x00000005
+      ]);
+      expect(apdu.getRawApdu()).toEqual(expectedApdu);
+    });
+  });
+
+  describe("parseResponse", () => {
+    it("should parse the response and extract the address", () => {
+      const responseData = Uint8Array.from("myAddressData", (c) =>
+        c.charCodeAt(0),
+      );
+      const response = new ApduResponse({
+        statusCode: Uint8Array.from([0x90, 0x00]),
+        data: responseData,
+      });
+
+      const result = command.parseResponse(response);
+
+      expect(result).toStrictEqual(
+        CommandResultFactory({
+          data: {
+            address: new TextDecoder().decode(responseData),
+          },
+        }),
+      );
+    });
+
+    it("should return an error if response status code is an error code", () => {
+      const errorStatusCode = Uint8Array.from([0x69, 0x85]);
+      const response = new ApduResponse({
+        statusCode: errorStatusCode,
+        data: new Uint8Array(),
+      });
+
+      const result = command.parseResponse(response);
+
+      expect(isSuccessCommandResult(result)).toBe(false);
+      if (!isSuccessCommandResult(result)) {
+        expect(result.error).toBeInstanceOf(BitcoinAppCommandError);
+        const error = result.error as BitcoinAppCommandError;
+        expect(error.customErrorCode).toBe("6985");
+        const expectedErrorInfo = bitcoinAppErrors["6985"];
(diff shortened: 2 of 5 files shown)


packages/signer/signer-btc/src/internal/app-binder/command/utils/bitcoinAppErrors.ts
```diff
@@ -0,0 +1,46 @@
...
+  "6D00": { message: "No command exists with the provided INS" },
+  "6E00": { message: "Bad CLA used for this application" },
+  B000: { message: "Wrong response length (buffer size problem)" },
+  B007: { message: "Aborted due to unexpected state reached" },
+  B008: { message: "Invalid signature or HMAC" },
+};
+
+export class BitcoinAppCommandError
+  extends DeviceExchangeError<void>
+  implements DmkError
```
jdabbech-ledger, line 31: `class  BitcoinAppCommandError extends DeviceExchangeError<BitcoinAppErrorCodes> { constructor(args: CommandErrorArgs<BitcoinAppErrorCodes>) { super(...) }}` should be enough, no need to add a customerrorcode imo

packages/signer/signer-btc/src/api/app-binder/GetWalletAddressCommandTypes.ts
```diff
@@ -0,0 +1,11 @@
+export type GetWalletAddressCommandResponse = {
+  readonly address: string;
+};
+
+export type GetWalletAddressCommandArgs = {
+  readonly display: 0 | 1;
```
paoun-ledger, line 6: [COULD] since it's an external API, I think types don't need to match the APDU, here it could just be a boolean to make it more clear

packages/signer/signer-btc/src/api/app-binder/GetWalletAddressCommandTypes.ts
```diff
@@ -0,0 +1,11 @@
...
+  readonly walletId: Buffer;
```
paoun-ledger, line 7: [COULD] We usually use UInt8Array everywhere instead of Buffer to make sure it's compatible with all environments, so it's not consistent with other apis to have buffers here

packages/signer/signer-btc/src/internal/app-binder/command/GetWalletAddressCommand.ts
```diff
@@ -0,0 +1,79 @@
...
+
+  getApdu(): Apdu {
+    const builder = new ApduBuilder({
+      cla: 0xe1,
+      ins: 0x03,
+      p1: 0x00,
+      p2: 0x00,
+    });
+
+    builder.addBufferToData(Buffer.from([this.args.display]));
```
paoun-ledger, line 41: [COULD] Same here you could just use `Uint8Array.from` instead of `Buffer.from`

.changeset/quiet-feet-leave.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ledgerhq/device-signer-kit-bitcoin": patch
+---
+
+Added GetWalletAddressCommand
```
jiyuzhuang, line 5: Present tense in change log ?

packages/signer/signer-btc/src/internal/app-binder/command/GetWalletAddressCommand.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export class GetWalletAddressCommand
+  implements
+    Command<GetWalletAddressCommandResponse, GetWalletAddressCommandArgs>
+{
+  args: GetWalletAddressCommandArgs;
+
+  constructor(args: GetWalletAddressCommandArgs) {
+    this.args = args;
+  }
```
jiyuzhuang, line 31:
```suggestion
  constructor(private readonly args: GetWalletAddressCommandArgs) {}
```
This will have the same effect and more clear, you can access to the property by `this.args` just as in the other places (most) in the code base.

packages/signer/signer-btc/src/internal/app-binder/command/GetWalletAddressCommand.ts
```diff
@@ -0,0 +1,79 @@
...
+    });
+
+    builder.addBufferToData(Buffer.from([this.args.display]));
+    builder.addBufferToData(this.args.walletId);
+    builder.addBufferToData(this.args.walletHmac);
+    builder.addBufferToData(Buffer.from([this.args.change]));
+    builder.add32BitUIntToData(this.args.addressIndex);
+
+    return builder.build();
+  }
```
jiyuzhuang, line 48: This APDU construction can be put in the chain as:
```ts
  return new ApduBuilder({...})
    .addXYZ(...)
    .addABC(...)
    .build();
```
This pattern should be taken if possible.

packages/signer/signer-btc/src/internal/app-binder/command/utils/bitcoinAppErrors.ts
```diff
@@ -0,0 +1,46 @@
...
+  | "B007"
+  | "B008";
+
+export const bitcoinAppErrors: CommandErrors<BitcoinAppErrorCodes> = {
+  "6985": { message: "Rejected by user" },
+  "6A86": { message: "Either P1 or P2 is incorrect" },
+  "6A87": { message: "Lc or minimum APDU length is incorrect" },
+  "6D00": { message: "No command exists with the provided INS" },
+  "6E00": { message: "Bad CLA used for this application" },
+  B000: { message: "Wrong response length (buffer size problem)" },
```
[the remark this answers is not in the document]
fAnselmi-Ledger: @paoun-ledger lint:)

packages/signer/signer-btc/src/api/app-binder/GetWalletAddressCommandTypes.ts
```diff
@@ -0,0 +1,11 @@
+export type GetWalletAddressCommandResponse = {
+  readonly address: string;
+};
+
+export type GetWalletAddressCommandArgs = {
```
paoun-ledger, line 5: [SHOULD] sorry I didn't notice at first, but signers commands should not be exposed in the api, it can remain internal. The commands cannot be easily used by end users since they require opening the app, waiting user interactions, using internal stuff such as merkletrees and tasks, ...

We already expose higher level apis such as `getAddress` of the `SignerBtc`, which will call that command internally

packages/signer/signer-btc/src/internal/app-binder/command/GetWalletAddressCommand.test.ts
```diff
@@ -0,0 +1,144 @@
...
+    const data = new Uint8Array([
+      args.display ? 1 : 0,
+      ...args.walletId,
+      ...args.walletHmac,
+      args.change ? 1 : 0,
+      (args.addressIndex >> 24) & 0xff,
+      (args.addressIndex >> 16) & 0xff,
+      (args.addressIndex >> 8) & 0xff,
+      args.addressIndex & 0xff,
+    ]);
```
paoun-ledger, line 41: [COULD] not sure what others think but for those kind of tests, I usually prefer to assert against a hardcoded buffer containing directly the expected apdu. Because here you almost reproduce the code of the function you're testing :sweat_smile: 
Something like that: https://github.com/LedgerHQ/device-sdk-ts/blob/develop/packages/signer/signer-eth/src/internal/app-binder/command/GetAddressCommand.test.ts#L9

packages/signer/signer-btc/src/api/app-binder/GetWalletAddressCommandTypes.ts
paoun-ledger: not yet resolved I think since that file was not removed ? :sweat_smile:

packages/signer/signer-btc/src/internal/app-binder/command/GetWalletAddressCommand.ts
```diff
@@ -0,0 +1,80 @@
...
+      .add32BitUIntToData(this.args.addressIndex)
+      .build();
+  }
+
+  parseResponse(
+    response: ApduResponse,
+  ): CommandResult<GetWalletAddressCommandResponse> {
+    const parser = new ApduParser(response);
+    const errorCode = parser.encodeToHexaString(response.statusCode);
+    if (isCommandErrorCode(errorCode, bitcoinAppErrors)) {
```
jdabbech-ledger, line 56: checking bitcoin error code is not enough, a `GLOBAL_ERROR` like `0x5515 (Device locked)` could happen here

packages/signer/signer-btc/src/api/app-binder/GetWalletAddressCommandTypes.ts
fAnselmi-Ledger: @paoun-ledger MB forgot to remove the file, thanks for checking in!

jdabbech-ledger (approved):

paoun-ledger (approved):
lgtm

packages/signer/signer-btc/src/internal/app-binder/command/GetWalletAddressCommand.ts
```diff
@@ -0,0 +1,88 @@
...
+    Command<GetWalletAddressCommandResponse, GetWalletAddressCommandArgs>
+{
+  constructor(private readonly args: GetWalletAddressCommandArgs) {}
+
+  getApdu(): Apdu {
+    return new ApduBuilder({
+      cla: 0xe1,
+      ins: 0x03,
+      p1: 0x00,
+      p2: 0x00,
```
paoun-ledger, line 43: [SHOULD] Actually after reading the reference implementation, I noticed p2 was set to 1 for all commands : https://github.com/LedgerHQ/app-bitcoin-new/blob/master/bitcoin_client_js/src/lib/appClient.ts#L89

And actually that was documented in the readme:

> The P1 field is reserved for future use and must be set to 0 in all messages. The P2 field is used as a protocol version identifier; the current version is 1, while version 0 is still supported. No other value must be used.

For now I think you can force support of version 1. If we need to support legacy apps, we may think about it later. In that case we should set p2=1 for all bitcoin commands :)

paoun-ledger (changes requested):
sorry I didn't notice previously that p2 should contain a protocol version...

paoun-ledger (approved):
lgtm

PR #604, opened 2025-01-13
🐛 (signer-eth) [DSDK-633]: Fallback on legacy signing

aussedatlo:
### 📝 Description
Update the signer ethereum device action to fallback on blind signing when some commands are not supported.

- `SignTransaction`: Fallback to blind sign transaction if getChallenge does not work
- `SignTypedObject`: Fallback to legacy signing if provide does not work.

### ❓ Context

- **JIRA or GitHub link**: [DSDK-633]

- **Feature**:

### ✅ Checklist

  - _list of the changes_

---

[DSDK-633]: https://ledgerhq.atlassian.net/browse/DSDK-633?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/signer/signer-eth/src/internal/app-binder/device-action/SignTransaction/SignTransactionDeviceAction.test.ts
+++ b/packages/signer/signer-eth/src/internal/app-binder/device-action/SignTransaction/SignTransactionDeviceAction.test.ts
@@ -4,6 +4,7 @@ import {
   DeviceActionStatus,
   hexaStringToBuffer,
   UnknownDAError,
+  UnknownDeviceExchangeError,
   UserInteractionRequired,
 } from "@ledgerhq/device-management-kit";
 import { InvalidStatusWordError } from "@ledgerhq/device-management-kit";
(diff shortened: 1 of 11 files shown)


packages/signer/signer-eth/src/internal/app-binder/device-action/SignTransaction/SignTransactionDeviceAction.test.ts
```diff
@@ -742,25 +765,86 @@ describe("SignTransactionDeviceAction", () => {
...
+          status: DeviceActionStatus.Completed,
         },
       ];
 
-      testDeviceActionStates(
+      const { observable } = testDeviceActionStates(
         deviceAction,
         expectedStates,
         makeDeviceActionInternalApiMock(),
         done,
```
jdabbech-ledger, line 811: [should] remove done from args and call it downside after obs complete expects, otherwise the test async is wrong

jdabbech-ledger (approved):

packages/signer/signer-eth/src/internal/app-binder/device-action/SignTransaction/SignTransactionDeviceAction.test.ts
```diff
@@ -178,7 +179,6 @@ describe("SignTransactionDeviceAction", () => {
         deviceAction,
         expectedStates,
         makeDeviceActionInternalApiMock(),
-        done,
```
ofreyssinet-ledger, line 181: [COMMENT] 🚨🚨🚨 I had not seen this alternative `testDeviceActionStates` where `done` is optional, it hasn't been introduced in this PR, but in my interpretation of it, it totally breaks it and renders it useless: if `done` is not passed to it and the observed states are wrong, nothing will happen no ?

aussedatlo: Removed 👍 we will open a new ticket to refactor this part.

jdabbech-ledger: The issue was that if we pass a wrong input in the device action, the call to testDeviceActionStates success and finish the corresponding test case even if the following expect fail.
Then if we run independently the test it succeeds (false positive), but if we run all the file tests the error is thrown in a wrong case by jest.
But you're right, removing done param here wasn't the correct solution and we've seen with @aussedatlo that we should rather create a ticket to refactor this. 
Will talk about it in CI

ofreyssinet-ledger (approved):

PR #656, opened 2025-02-03
✨ (signer-eth) [DSDK-652]: Add web3checks support

aussedatlo:
### 📝 Description

Add web3checks support:
- create a web3checks loader in the `context-module`
- add a `web3checks` step in the `signTransactionDA`
- add the web3checks provider `url` param available to edit in the `sample`

### ❓ Context

- **JIRA or GitHub link**:

- **Feature**:

### ✅ Checklist

  - _list of the changes_

---

### DIFF
--- a/packages/signer/context-module/src/pki/model/PkiCertificateInfo.ts
+++ b/packages/signer/context-module/src/pki/model/PkiCertificateInfo.ts
@@ -1,9 +1,7 @@
 import type { KeyUsage } from "@/pki/model/KeyUsage";
 
-import { type KeyId } from "./KeyId";
-
 export type PkiCertificateInfo = {
   targetDevice: string;
   keyUsage: KeyUsage;
-  keyId: KeyId;
+  keyId: string;
 };
--- a/packages/signer/context-module/src/pki/model/PkiCertificateInfo.ts
+++ b/packages/signer/context-module/src/pki/model/PkiCertificateInfo.ts
@@ -1,7 +1,9 @@
 import type { KeyUsage } from "@/pki/model/KeyUsage";
 
+import { type KeyId } from "./KeyId";
+
 export type PkiCertificateInfo = {
   targetDevice: string;
   keyUsage: KeyUsage;
-  keyId?: string | undefined;
+  keyId: KeyId;
 };
--- a/packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
+++ b/packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
@@ -1,0 +1,82 @@
+import axios from "axios";
+import { inject, injectable } from "inversify";
+import { Either, Left, Right } from "purify-ts";
+
+import { configTypes } from "@/config/di/configTypes";
+import type { ContextModuleConfig } from "@/config/model/ContextModuleConfig";
+import {
+  type Web3CheckContext,
+  type Web3Checks,
+} from "@/web3-check/domain/web3CheckTypes";
+import PACKAGE from "@root/package.json";
+
+import { Web3CheckDataSource } from "./Web3CheckDataSource";
+import { GetWeb3ChecksRequestDto, Web3CheckDto } from "./Web3CheckDto";
+
+@injectable()
+export class HttpWeb3CheckDataSource implements Web3CheckDataSource {
+  constructor(
+    @inject(configTypes.Config) private readonly config: ContextModuleConfig,
+  ) {}
+
+  async getWeb3Checks(
+    params: Web3CheckContext,
+  ): Promise<Either<Error, Web3Checks>> {
+    let web3CheckDto: Web3CheckDto;
+
+    try {
+      const requestDto: GetWeb3ChecksRequestDto = {
+        tx: {
+          from: params.from,
+          raw: params.rawTx,
+        },
+        chain: params.chainId,
+        preset: "blockaid",
+      };
+      const response = await axios.request<Web3CheckDto>({
+        method: "POST",
+        url: `${this.config.web3checks.url}/ethereum/scan/tx`,
+        data: requestDto,
+        headers: {
+          "X-Ledger-Client-Version": `context-module/${PACKAGE.version}`,
+        },
+      });
+
+      web3CheckDto = response.data;
+    } catch (_error) {
+      return Left(
+        new Error(
+          "[ContextModule] HttpWeb3CheckDataSource: Failed to fetch web3 checks informations",
+        ),
+      );
+    }
+
+    if (!this.isWeb3CheckDto(web3CheckDto)) {
+      return Left(
+        new Error(
+          "[ContextModule] HttpWeb3CheckDataSource: Cannot exploit Web3 checks data received",
+        ),
+      );
+    }
+
+    const result: Web3Checks = {
+      publicKeyId: web3CheckDto.public_key_id,
+      descriptor: web3CheckDto.descriptor,
+    };
+
+    return Right(result);
+  }
+
+  private isWeb3CheckDto(dto: unknown): dto is Web3CheckDto {
+    return (
+      dto != null &&
+      typeof dto == "object" &&
+      "public_key_id" in dto &&
+      dto.public_key_id != null &&
+      typeof dto.public_key_id == "string" &&
+      "descriptor" in dto &&
+      dto.descriptor != null &&
+      typeof dto.descriptor == "string"
+    );
+  }
+}
--- a/packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
+++ b/packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
@@ -68,7 +68,7 @@ export class HttpWeb3CheckDataSource implements Web3CheckDataSource {
 
     const certificate = await this._certificateLoader.loadCertificate({
       keyId: web3CheckDto.public_key_id,
-      keyUsage: "replace-me" as KeyUsage, // TODO: replace with the keyUsage given by the API
+      keyUsage: KeyUsage.TxSimulationSigner,
       targetDevice: deviceModelId,
     });
 
--- a/packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
+++ b/packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
@@ -10,6 +10,7 @@ import { KeyUsage } from "@/pki/model/KeyUsage";
 import {
   type Web3CheckContext,
   type Web3Checks,
+  type Web3CheckTypedDataContext,
 } from "@/web3-check/domain/web3CheckTypes";
 import PACKAGE from "@root/package.json";
 
@@ -31,7 +32,7 @@ export class HttpWeb3CheckDataSource implements Web3CheckDataSource {
     let web3CheckDto: Web3CheckDto;
     let requestDto: GetWeb3ChecksRequestDto;
     let url: string;
-    if ("data" in context) {
+    if (this.isTypedDataContext(context)) {
       requestDto = {
         msg: {
           from,
@@ -89,6 +90,12 @@ export class HttpWeb3CheckDataSource implements Web3CheckDataSource {
     };
 
     return Right(result);
+  }
+
+  private isTypedDataContext(
+    context: Web3CheckContext,
+  ): context is Web3CheckTypedDataContext {
+    return "data" in context;
   }
 
   private isWeb3CheckDto(dto: unknown): dto is Web3CheckDto {
--- a/packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
+++ b/packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
@@ -4,6 +4,9 @@ import { Either, Left, Right } from "purify-ts";
 
 import { configTypes } from "@/config/di/configTypes";
 import type { ContextModuleConfig } from "@/config/model/ContextModuleConfig";
+import { pkiTypes } from "@/pki/di/pkiTypes";
+import { type PkiCertificateLoader } from "@/pki/domain/PkiCertificateLoader";
+import { KeyUsage } from "@/pki/model/KeyUsage";
 import {
   type Web3CheckContext,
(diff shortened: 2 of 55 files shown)


packages/signer/signer-eth/src/internal/app-binder/task/GetWeb3CheckTask.ts
```diff
@@ -0,0 +1,104 @@
...
+    private readonly api: InternalApi,
+    private readonly args: GetWeb3CheckTaskArgs,
+  ) {}
+
+  async run(): Promise<GetWeb3CheckTaskResult> {
+    const { contextModule, mapper, transaction } = this.args;
+    const { deviceModelId } = this.api.getDeviceSessionState();
+
+    const parsed = mapper.mapTransactionToSubset(transaction);
+    parsed.ifLeft((err) => {
```
valpinkman, line 46: [COULD] Same here, you could leverage the Either map functions

```typescript
const { subset, serializedTransaction } = mapper.mapTransactionToSubset(transaction).map(parsed => parsed).ifLeft((err) => {
  throw err;
})
```
I feel it's better than using `unsafeCoerce`. WDYT ?

packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
```diff
@@ -0,0 +1,97 @@
...
+
+    if (!this.isWeb3CheckDto(web3CheckDto)) {
+      return Left(
+        new Error(
+          "[ContextModule] HttpWeb3CheckDataSource: Cannot exploit Web3 checks data received",
+        ),
+      );
+    }
+
+    const certificate = await this._certificateLoader.loadCertificate({
```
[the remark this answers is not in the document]
aussedatlo: This method may fail, but explicit result checking is unnecessary as failure is handled later

packages/signer/context-module/src/web3-check/domain/DefaultWeb3CheckLoader.ts
```diff
@@ -0,0 +1,45 @@
...
+  constructor(
+    @inject(web3CheckTypes.Web3CheckDataSource)
+    dataSource: Web3CheckDataSource,
+  ) {
+    this._dataSource = dataSource;
+  }
+
+  async load(
+    web3CheckContext: Web3CheckContext,
+  ): Promise<Either<Error, Web3Checks>> {
```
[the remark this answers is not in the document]
aussedatlo: These errors are used only internally, what do you suggest ?

packages/signer/signer-eth/src/internal/app-binder/task/GetWeb3CheckTask.ts
aussedatlo: Throwing on tasks is safe since the device action catches it. this method avoid using `unsafeCoerce` and is an equivalent so can add it 👍

packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
```diff
@@ -0,0 +1,97 @@
...
+    let web3CheckDto: Web3CheckDto;
+
+    try {
+      const requestDto: GetWeb3ChecksRequestDto = {
+        tx: {
+          from,
+          raw: rawTx,
+        },
+        chain: chainId,
+        preset: "blockaid",
```
paoun-ledger, line 42: [SHOULD] preset can be removed as it's not necessarily blockaid anymore

aussedatlo: Should I need to remove `blockaid` from `KeyId` too ?

packages/signer/context-module/src/pki/model/PkiCertificateInfo.ts
```diff
@@ -1,7 +1,9 @@
 import type { KeyUsage } from "@/pki/model/KeyUsage";
 
+import { type KeyId } from "./KeyId";
+
 export type PkiCertificateInfo = {
   targetDevice: string;
   keyUsage: KeyUsage;
-  keyId?: string | undefined;
+  keyId: KeyId;
```
paoun-ledger, line 8: [ASK] I don't know if a strong typing is good here. The signer kit should care about the use case of the certificate (KeyUsage), but not necessarily the entity computing the signature (KeyId). Most of the time, we will retrieve the key id of a descriptor from the CAL as a string. For example  the signer kit has to know the cetificate usage if for `tx_simulation`. But it probably don't care if the current partner is `blockaid`, `tunderly` or any other one.
It also mean that if Ledger add or remove a provider, we may not want having to release a new signer kit for that.
What do you think?

packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
```diff
@@ -0,0 +1,97 @@
...
+      return Left(
+        new Error(
+          "[ContextModule] HttpWeb3CheckDataSource: Cannot exploit Web3 checks data received",
+        ),
+      );
+    }
+
+    const certificate = await this._certificateLoader.loadCertificate({
+      keyId: web3CheckDto.public_key_id,
+      keyUsage: "replace-me" as KeyUsage, // TODO: replace with the keyUsage given by the API
```
paoun-ledger, line 72: [SHOULD] actually the usage is known, it's tx_simulation. Or did I miss something?

packages/signer/signer-eth/src/internal/app-binder/task/ProvideTransactionFieldDescriptionTask.ts
```diff
@@ -329,6 +330,10 @@ export class ProvideTransactionFieldDescriptionTask {
             `The context type [${type}] is not valid as a transaction field or metadata`,
           ),
         });
+      case ClearSignContextType.WEB3_CHECK:
```
paoun-ledger, line 333: [SHOULD] You should probably provide the certificate in that case, as you did in `ProvideTransactionContextTask`? Or else this command won't work (there is no fallback key for web3checks)

packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
paoun-ledger: mmm not sure but I'd say yes. In my opinion we can keep Ledger keys there, but not partners keys if they can be retrieved dynamically from the CAL. Or else we will have to maintain that enum content when we change partners list

packages/signer/signer-eth/src/internal/app-binder/task/ProvideTransactionFieldDescriptionTask.ts
aussedatlo: we already load the certificate on line `73` at the beginning of the `run` function

packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
aussedatlo: Ok will replace it 👍

packages/signer/context-module/src/pki/model/PkiCertificateInfo.ts
aussedatlo: I agree, we can just accept a `string` here

packages/signer/signer-eth/src/internal/app-binder/task/GetWeb3CheckTask.ts
```diff
@@ -0,0 +1,104 @@
...
+import { type TransactionMapperService } from "@internal/transaction/service/mapper/TransactionMapperService";
+
+export type GetWeb3CheckTaskResult =
+  | {
+      readonly web3Check: ClearSignContextSuccess<ClearSignContextType.WEB3_CHECK> | null;
+    }
+  | {
+      readonly web3Check: null;
+      error: DmkError;
+    };
```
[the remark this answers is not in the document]
paoun-ledger: I was thinking of getting rid of `GetWeb3CheckTaskResult`, especially `error: DmkError` which is a mapping of device interactions error (getAddress, getAppConfig). But fell free to ignore, actually it also make sense to have a custom return type for that kind of task

packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
paoun-ledger: it's `tx_simu_signer` indeed

packages/signer/context-module/src/web3-check/data/HttpWeb3CheckDataSource.ts
```diff
@@ -24,31 +24,41 @@ export class HttpWeb3CheckDataSource implements Web3CheckDataSource {
...
+    context: Web3CheckContext,
+  ): Promise<Either<Error, Web3Checks>> {
+    const { from, deviceModelId } = context;
     let web3CheckDto: Web3CheckDto;
-
-    try {
-      const requestDto: GetWeb3ChecksRequestDto = {
+    let requestDto: GetWeb3ChecksRequestDto;
+    let url: string;
+    if ("data" in context) {
```
aussedatlo, line 34: [COULD]: use a typeguard for clarity

packages/signer/context-module/src/web3-check/data/Web3CheckDto.ts
```diff
@@ -1,13 +1,21 @@
-export type GetWeb3ChecksRequestDto = {
-  tx: {
-    from: string;
-    raw: string;
-  };
-  chain: number;
-};
+import type { Web3CheckTypedData } from "@/web3-check/domain/web3CheckTypes";
+
+export type GetWeb3ChecksRequestDto =
```
aussedatlo, line 3: [COULD] create sub types, for example:
```ts
export type GetWeb3ChecksRequestDto =
  | GetWeb3ChecksTypedDataRequestDto
  | GetWeb3ChecksRawTxRequestDto;

export type GetWeb3ChecksTypedDataRequestDto = {
  msg: {
    from: string;
    data: Web3CheckTypedData;
  };
};

export type GetWeb3ChecksRawTxRequestDto = {
  tx: {
    from: string;
    raw: string;
  };
  chain: number;
};
```

totally not required but I think it can be easier to understand

packages/signer/context-module/src/web3-check/domain/web3CheckTypes.ts
```diff
@@ -2,13 +2,32 @@ import { type DeviceModelId } from "@ledgerhq/device-management-kit";
...
+    chainId?: number;
+    verifyingContract?: string;
+    salt?: string;
+  };
+  types: Record<string, Array<{ name: string; type: string }>>;
+  primaryType: string;
+  message: Record<string, unknown>;
 };
 
+export type Web3CheckContext =
```
aussedatlo, line 18: [COULD] same here

packages/signer/signer-eth/src/internal/app-binder/task/GetWeb3CheckTask.ts
```diff
@@ -79,14 +78,35 @@ export class GetWeb3CheckTask {
...
+
+    if ("transaction" in this.args) {
+      // Transaction simulation
+      const parsed = this.args.mapper.mapTransactionToSubset(
+        this.args.transaction,
+      );
+      parsed.ifLeft((err) => {
+        throw err;
+      });
+      const { subset, serializedTransaction } = parsed.unsafeCoerce();
```
aussedatlo, line 109: [COULD] replace 'unsafeCoerce' with 'map' or 'caseOf' or something else.
It's easy to read as it avoid another block condition but still remains unsafe, wdyt ?

packages/signer/signer-eth/src/internal/app-binder/task/ProvideEIP712ContextTask.ts
```diff
@@ -65,6 +72,21 @@ export class ProvideEIP712ContextTask {
   ) {}
 
   async run(): ProvideEIP712ContextTaskReturnType {
+    // Send message simulation first
+    if (this.args.web3Check) {
+      if (this.args.web3Check.certificate) {
```
aussedatlo, line 79: [ASK] It is possible to have a `web3Check` without `certificate` that work on the device ?

packages/signer/signer-eth/src/internal/app-binder/task/GetWeb3CheckTask.ts
```diff
@@ -46,17 +47,32 @@ export class GetWeb3CheckTask {
...
     if (!isSuccessCommandResult(configResult)) {
       return {
         web3Check: null,
         error: configResult.error,
       };
     }
+    let web3ChecksEnabled = configResult.data.web3ChecksEnabled;
+
+    // If feature is disabled and opt-in was not done, trigger it on the device
+    if (!web3ChecksEnabled && !configResult.data.web3ChecksOptIn) {
```
aussedatlo, line 61: [ASK] what happen for a device without web3checks support ? I assume that `web3ChecksOptIn` would be `false`, but 'Web3CheckOptInCommand' not supported in the device? Should we add an app version check here to enable this feature only for 1.15.0+ apps?

aussedatlo (commented):
Nice work 🚀 , two questions and few optional comments.

packages/signer/signer-eth/src/internal/app-binder/task/ProvideEIP712ContextTask.ts
paoun-ledger: not possible, but that's a kind of implementation detail on device side. On some descriptors the certificate is optional (with a fallback in the device), so I prefer to stay agnostic of device actual implementation here. WDYT?

aussedatlo: Ok for me 👍

aussedatlo:
LGTM

packages/signer/context-module/src/web3-check/domain/DefaultWeb3CheckLoader.ts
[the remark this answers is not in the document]
valpinkman: fair point if internal only

valpinkman (approved):

PR #685, opened 2025-02-17
✨ (dmk) [NO-ISSUE]: Add genuine check device action

jiyuzhuang:
#### Demo:

https://github.com/user-attachments/assets/44820c3c-9679-4359-ab50-00cce9fd2a32

#### Notes:
- This PR includes the task where the WebSocket message events have been handled and the **Genuine Check** device action;
- The latest **3** events emitted in the observable have been stored in the `intermediateValue` to help us tracking the process (in this device action) and to inform end user the process in coming ones;
- To test the `AllowSecureConnection` state, you need to reboot your device after each test in order to make it happen because this permission will only be requested once before reboot.

### ❓ Context

### ✅ Checklist

  - Genuine check in DMK

---


packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.test.ts
```diff
@@ -0,0 +1,269 @@
...
+        nonce: 1,
+        data: "e051000000",
+      }),
+      type: "",
+      target: {} as WebSocket,
+    });
+
+    vi.advanceTimersByTime(500);
+
+    await new Promise((resolve) => setTimeout(resolve, TEST_DELAY));
```
valpinkman, line 114: [ASK] Why do you need a timeout if you advenceTimers ?

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.ts
```diff
@@ -0,0 +1,343 @@
...
+                return;
+              }
+
+              response.caseOf({
+                Left: (error) => {
+                  subscriber.error(new SecureChannelError(error));
+                },
+                Right: (apduResponse: ApduResponse) => {
+                  let outMessageResponse: OutMessageResponseEnum;
+                  /**
```
valpinkman, line 171: 🤩

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.ts
```diff
@@ -0,0 +1,343 @@
...
+    } else {
+      throw new SecureChannelError(
+        `Invalid WebSocket connection: ${String(this._args.connection.extract())}`,
+      );
+    }
+  }
+
+  run(): Observable<SecureChannelEvent> {
+    const obs = new Observable<SecureChannelEvent>((subscriber) => {
+      let unsubscirbed: boolean = false;
```
valpinkman, line 52: [SHOULD] Fix typo here, should be `unsubscribed`

valpinkman (changes requested):
Pretty good, just some comments and a small typo

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.test.ts
jiyuzhuang: The advancing time is to make the mock `sendApduFn` return a value ASAP (1ms after advancing), then wait for 5ms to make sure that it's already responded.

```typescript
    sendApduFn.mockImplementationOnce(
      () =>
        new Promise((r) =>
          setTimeout(
            () =>
              r(
                Right({
                  data: new Uint8Array([0x90, 0x00]),
                  statusCode: new Uint8Array([0x90, 0x00]),
                }),
              ),
            500 + 1,
          ),
        ),
    );
```

valpinkman (approved):

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.test.ts
valpinkman: [ASK] Why not advance timers by 1000s instead ?

packages/device-management-kit/src/api/secure-channel/utils.ts
```diff
@@ -0,0 +1,43 @@
...
+ * Checks if the user refused the permission on the device.
+ *
+ * The following status codes indicate a refusal:
+ * - "0x5501": User refused the permission.
+ * - "0x6985": Condition of use not satisfied.
+ *
+ * @param statusCode - The status code to check.
+ * @returns `true` if the status code indicates a refusal, otherwise `false`.
+ */
+export function isRefusedByUser(statusCode: Uint8Array): boolean {
```
fAnselmi-Ledger, line 26: [COULD] can this be part of CommandUtils?

packages/device-management-kit/src/api/secure-channel/device-action/GenuineCheck/GenuineCheckDeviceAction.ts
```diff
@@ -0,0 +1,429 @@
...
+                    ..._.context.intermediateValue,
+                    history,
+                  };
+                },
+                _internalState: (_) => {
+                  if (
+                    _.event.snapshot.context?.type ===
+                    SecureChannelEventType.Result
+                  ) {
+                    console.log(_.event.snapshot.context);
```
jdabbech-ledger, line 329: [should] remove log 👀

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.ts
```diff
@@ -0,0 +1,343 @@
...
+import { type Either } from "purify-ts";
+import { Observable } from "rxjs";
+
+import {
+  type ApduResponse,
+  bufferToHexaString,
+  CommandUtils,
+  hexaStringToBuffer,
+  type InternalApi,
+} from "@api/index";
```
jdabbech-ledger, line 11: [should] import from complete path instead of `api/index`

packages/device-management-kit/src/api/secure-channel/utils.ts
jiyuzhuang: We do not have command here, this is a send APDU response checker, about the success check, because it already exists. In addition, the criteria is just for this response, it includes the normal refuse code `0x5501` and condition not satisfied `0x6985`, so a specific treatment.

packages/device-management-kit/src/api/secure-channel/device-action/GenuineCheck/GenuineCheckDeviceAction.ts
```diff
@@ -0,0 +1,429 @@
...
+    };
+
+    const genuineCheck = ({
+      input,
+    }: Input<{
+      deviceInfo: GetOsVersionResponse;
+      finalFirmware: FinalFirmware;
+    }>) => {
+      const { deviceInfo, finalFirmware } = input;
+      const eitherConnection = internalApi
```
jdabbech-ledger, line 414: [ask]  As internalApi is injected, why creating the connection in the device action instead of in the task?

jiyuzhuang: The idea is to isolate the WebSocket events handlers with the connections and make it generic. If we try to get the connection inside the task, we will add a parameter in the constructor to indicate which connection we want. The actual solution is like: (consumer) prepare the connection, then pass it to me (task), I will handle it and I don't care what the connection is, I will handle them in the same way.

packages/device-management-kit/src/api/secure-channel/utils.ts
fAnselmi-Ledger: Gotcha! wasn't sure abt 0x6985, left the msg as you used it together with CommandUtils.isLockedDeviceResponse:)

packages/device-management-kit/src/api/secure-channel/device-action/GenuineCheck/types.ts
```diff
@@ -0,0 +1,81 @@
...
+
+export type Input<T> = { input: T };
+
+export type GenuineCheckDAOutput = { isGenuine: boolean };
+
+export type GenuineCheckDAInput = GoToDashboardDAInput;
+
+export type GenuineCheckDAError =
+  | CommandErrorResult["error"]
+  | DeviceLockedError
```
jdabbech-ledger, line 32: [should] this error should be handled by command error result

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.ts
```diff
@@ -0,0 +1,340 @@
...
+export class ConnectToSecureChannelTask {
+  private readonly _connection: WebSocket;
+  constructor(
+    private readonly _api: InternalApi,
+    private readonly _args: ConnectToSecureChannelTaskArgs,
+  ) {
+    if (this._args.connection.isRight()) {
+      this._connection = this._args.connection.extract();
+    } else {
+      throw new SecureChannelError(
```
jdabbech-ledger, line 41: [could] this error could not be parsed well in client side, as it's handled by `assignErrorFromEvent` in DA. IMO it should be handled in a better way if the check is done within the `run` observable

packages/device-management-kit/src/api/secure-channel/device-action/GenuineCheck/types.ts
jiyuzhuang: Ah, yes, and I think we need to update this in all device actions.

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.test.ts
```diff
@@ -0,0 +1,269 @@
...
+    const obs = task.run();
+    obs.subscribe((e) => events.push(e));
+
+    expect(mockWebSocket.onopen).toBeDefined();
+    mockWebSocket.onopen!({
+      type: "open",
+      target: {} as WebSocket,
+    });
+
+    await new Promise((resolve) => setTimeout(resolve, TEST_DELAY));
```
ofreyssinet-ledger, line 70: [ASK] would this work instead with await `new Promise(setImmediate)` ? (same question for every test that uses that)

ofreyssinet-ledger (changes requested):
Nice 👍
I think the "connect to secure channel task" is more than a task a we defined it: it's not very "linear" and it would be nice to think about a state machine implementation for it, it might be a bit simpler to read, and also simpler to document. Then it would make sense to add some documentation with diagrams, so it's very easy to understand for everyone.
Just a few small comments, and I think the arbitrary timeout added in the task is not needed (not at this level) so it should be removed.

packages/device-management-kit/src/api/secure-channel/device-action/GenuineCheck/GenuineCheckDeviceAction.ts
```diff
@@ -0,0 +1,428 @@
...
+          return Left(context._internalState.error);
+        } else {
+          return Right(context._internalState.result);
+        }
+      },
+    });
+  }
+
+  private extractDependencies(internalApi: InternalApi): MachineDependencies {
+    const provider = 1; // TODO: get the provider from user configuration
```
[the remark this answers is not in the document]
jiyuzhuang: IDK, this is not only our side decision, will be discussed later.

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.test.ts
jiyuzhuang: No, it won't work. Here what we need is a real waiting. Putting a immediate resolved promise at the end of event loop queue has nothing to do with it.

packages/device-management-kit/src/api/secure-channel/device-action/GenuineCheck/GenuineCheckDeviceAction.ts
```diff
@@ -0,0 +1,428 @@
...
+                    SecureChannelEventType.PermissionGranted
+                  )
+                    return {
+                      history,
+                      requiredUserInteraction: UserInteractionRequired.None,
+                    };
+                  return {
+                    ..._.context.intermediateValue,
+                    history,
+                  };
```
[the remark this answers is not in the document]
jiyuzhuang: It seems like a good idea, I thought there were only 2 out of 9 events need to be treated particularly, so they are particularities of all.

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.ts
```diff
@@ -0,0 +1,340 @@
...
+              if (willRequestPermission(apdu)) {
+                waitingForUserAction = true;
+                userActionTimeout = setTimeout(() => {
+                  if (unsubscribed) {
+                    return;
+                  }
+                  subscriber.next({
+                    type: SecureChannelEventType.PermissionRequested,
+                  });
+                }, USER_ACTION_TIMEOUT);
```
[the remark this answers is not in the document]
jiyuzhuang: The `SecureChannelEventType.PermissionRequested` event is forecasting, we got this info before sending the APDU, adding the timeout is trying to synchronise the event emitted in the observable and the event taken place on the device. I don't totally understand the comment in LL, this event will take place only one time every rebooting.

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.ts
```diff
@@ -0,0 +1,340 @@
...
+  type WebSocketConnectionError,
+} from "@internal/secure-channel/model/Errors";
+
+const USER_ACTION_TIMEOUT = 500;
+
+export type ConnectToSecureChannelTaskArgs = {
+  connection: Either<WebSocketConnectionError, WebSocket>;
+};
+
+export class ConnectToSecureChannelTask {
```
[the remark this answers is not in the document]
jiyuzhuang: Study subject in the future, maybe in a tech day. I'm not sure it can be done within a state machine as we don't need to care about the sequence of received message.

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.ts
jiyuzhuang: You're right, I thinking about it.

packages/device-management-kit/src/api/secure-channel/device-action/GenuineCheck/types.ts
```diff
@@ -0,0 +1,77 @@
...
+  | HttpFetchApiError
+  | UnknownDAError;
+
+export type GenuineCheckDARequiredInteraction =
+  | UserInteractionRequired.None
+  | UserInteractionRequired.UnlockDevice
+  | UserInteractionRequired.AllowSecureConnection;
+
+export type GenuineCheckDAIntermediateValue = {
+  history: SecureChannelEvent[];
```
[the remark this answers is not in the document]
jiyuzhuang: I agree with you, this was not here firstly. Then I thought it may be better to add it for demonstrating and tracking objectifs. If not, we will just got a `true` or `false` when the job is done without any visibilities of the process. We will expose the least data to the end user before releasing.

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.ts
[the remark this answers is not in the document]
ofreyssinet-ledger: It's forecasting indeed, but it's the same for every "user interaction required" that we implement whenever we send a command like open app or sign etc.
The delay in LL was added so that the state right before this one doesn't appear for less that 500ms. But I really don't think it's required here, and not having it would simplify the implem.

packages/device-management-kit/src/api/secure-channel/device-action/GenuineCheck/types.ts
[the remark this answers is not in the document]
ofreyssinet-ledger: A history is not needed because the consumer gets an observable, with all the previous values.
And to have less data you could perhaps just show the event type:
```
lastEventType: SecureChannelEventType,
```
Just a suggestion ...

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.ts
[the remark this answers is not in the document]
ofreyssinet-ledger: The sequence is important because for instance once your in "bulk mode" you ignore other events from the server (and also socket disconnection).
I think those variables are signs that we could modelize this easily with a state machine:
```
inBulkMode: boolean,
communicationFinished: boolean,
waitingForUserAction: boolean,
```

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.ts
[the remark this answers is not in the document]
ofreyssinet-ledger: Also the fact that the permission request will appear only once on the device is something that we can anticipate: in the result of the getOsVersion command, there are some flags that indicate if the user has already allowed a connection with the Ledger HSM.
We should use that information if we want to avoid emitting a wrong `SecureChannelEventType.PermissionRequested` event. It will be much better than this timeout.

https://ledgerhq.atlassian.net/wiki/spaces/FW/pages/4455596105/Dashboard+APDU+commands#SE-flags

packages/device-management-kit/src/api/secure-channel/device-action/GenuineCheck/GenuineCheckDeviceAction.ts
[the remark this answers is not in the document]
jiyuzhuang: Already updated yesterday ✨

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.ts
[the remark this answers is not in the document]
jiyuzhuang: Is the `DeviceSessionStateType.ReadyWithSecureChannel` designed for it ?

packages/device-management-kit/src/api/secure-channel/device-action/GenuineCheck/GenuineCheckDeviceAction.ts
```diff
@@ -0,0 +1,428 @@
...
+              target: "Success",
+            },
+          ],
+        },
+        Success: {
+          type: "final",
+        },
+        Error: {
+          type: "final",
+        },
```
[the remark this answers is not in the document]
jiyuzhuang: It seems like other device actions have the same state definitions.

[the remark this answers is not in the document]
ofreyssinet-ledger: Yep let's keep it like that it's fine

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.ts
```diff
@@ -0,0 +1,346 @@
...
+   * @returns {boolean} `true` if a secure connection is allowed, otherwise `false`.
+   */
+  isSecureConnectionAllowed(): boolean {
+    const deviceSessionState = this._api.getDeviceSessionState();
+    if (
+      deviceSessionState.sessionStateType !== DeviceSessionStateType.Connected
+    ) {
+      return !!deviceSessionState.deviceState?.isSecureConnectionAllowed;
+    }
+    return false;
```
ofreyssinet-ledger, line 344: [COULD]

`return ("isSecureConnectionAllowed" in deviceSessionState.deviceState) && deviceSessionState.isSecureConnectionAllowed`

ofreyssinet-ledger: so you don't have to check `sessionStateType`

packages/device-management-kit/src/api/device-session/types.ts
```diff
@@ -2,3 +2,57 @@
...
+};
+
+// 2nd byte of the secure element flags, placeholder for the endorsement information
+export type EndorsementInformation = unknown;
+
+// 3rd byte of the secure element flags, placeholder for the words information
+export type WordsInformation = unknown;
+
+// 4th byte of the secure element flags, placeholder for the onboarding status
+export type OnboardingStatus = unknown;
```
ofreyssinet-ledger, line 58: Move this to a new file in `api/device/`

ofreyssinet-ledger (changes requested):
Overall it's better but the parsing of the flags should be done at the command parsing level, so it doesn't have to be done again anywhere else.

packages/device-management-kit/src/api/secure-channel/constants.ts
```diff
@@ -0,0 +1 @@
+export const GENUINE_DEVICE = "0000";
```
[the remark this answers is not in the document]
jiyuzhuang: For me, they are the same thing and I prefer shorter naming if this is no confusion. `event.payload === GENUINE_DEVICE_RESULT` is less obvious than `=== GENUINE_DEVICE` which tells this is a genuine device. Besides, we don't care about other cases (not genuine), so no other result comparing to genuine device.

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.ts
jiyuzhuang: And I think there are also other places need to be updated.

packages/device-management-kit/src/api/secure-channel/constants.ts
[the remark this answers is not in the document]
jiyuzhuang: But I will rename to let this go ;-)

packages/device-management-kit/src/api/secure-channel/utils.ts
```diff
@@ -0,0 +1,116 @@
...
+
+  /**
+   * Retrieves the onboarding status based on the secure element flags, which is the fourth byte of the flags.
+   * @returns {OnboardingStatus}
+   * @throws Will throw an error if the method is not implemented.
+   */
+  onboardingStatus(): OnboardingStatus {
+    throw new Error("Not implemented");
+  }
+}
```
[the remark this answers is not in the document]
ofreyssinet-ledger: We can keep the raw flags, but they should be parsed with the rest of the command result. It's not "secondary" or anything like that it's just that the data is encapsulated in a different way. Just do all the parsing in the same place for the sake of simplicity. The raw flags don't mean anything that is usable without getting parsed.

If it deserves a better place, then it should be done right away in the PR that introduces, not "later", why would we do it later and not now ? The more we do that, the more messy the DMK will get. We should try to keep it as tidy as possible as we work on it, because we will never allocate time later on just for cleaning up small things like that. Let's try to do things the right way on the first try, not just push back as an other task for later.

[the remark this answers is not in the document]
jiyuzhuang: These four `deviceGeneralState`,  `endorsementInformation`, `wordsInformation` and `onboardingStatus` object will be in the same level, the separate definitions are dedicated to have a better visibility and make it easy to read. In addition to my previous comment, I would like to keep the parsed APDU response is relavant to [this table](https://ledgerhq.atlassian.net/wiki/spaces/FW/pages/4455596105/Dashboard+APDU+commands#1.1.-Get-version[inlineExtension]).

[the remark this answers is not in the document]
ofreyssinet-ledger: in `GetOsVersionCommandResult`, you can have

```
type GetOsVersionCommandResult = {
  /*...*/
  deviceGeneralState: DeviceGeneralState;
  endorsementInformation: EndorsementInformation;
  wordsInformation: WordsInformation;
  onboardingStatus: OnboardingStatus;
}
```

What's the issue with that? And if you don't want to implement the parsing of `EndorsementInformation`, `WordsInformation`, `OnboardingStatus` for now because they are not needed, then just ignore completely those for now.

packages/device-management-kit/src/api/secure-channel/constants.ts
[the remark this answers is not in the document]
ofreyssinet-ledger: Thanks 👍

packages/device-management-kit/src/api/secure-channel/utils.ts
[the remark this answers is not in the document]
ofreyssinet-ledger: The goal of parsing the result of a command is to extract all the information contained in the response. Whether it's encoded in a length-value way, of if it's encoded in flags or whatever. That way the rest of the DMK, or the client, can use that information without needing to know the specifics of how the data is encoded for that command.

packages/device-management-kit/src/api/secure-channel/task/ConnectToSecureChannelTask.ts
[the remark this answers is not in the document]
ofreyssinet-ledger: 1. GetOSVersion parses the value
2. in your GenuineCheck device action, you take that value and store it in the device state
3. exactly what you do right now, in the task you get the current device state

packages/device-management-kit/src/api/secure-channel/utils.ts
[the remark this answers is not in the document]
jiyuzhuang: I think you misunderstand my comment. I mean later is to say, today this is a proper place for it, if tomorrow we introduce some other use case of this command, we relocate it if needed. For example, this information (`seFlags`) is not always needed when sending this command, if so, we don't need to parse it.
And about the structure, we will not have the four objects separately because they come from the same field. I define them in four different type is to improve the readability and reference to the 4 bytes, but you should see that I use `&` to combine them together, so we will have:
```
secureElementFlags: {
  // First byte
  isXXXX: boolean;
  isYYYY: boolean;
  // Second byte
  someThing: unknown;
  ...
} 
```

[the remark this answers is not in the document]
jiyuzhuang: In fact, I don't want the device session state 'unstable', sometimes it has some property, sometimes doesn't. The device session state will not have those properties from the response of GetOsVersionCommand including all the flags if this command isn't invoked. So every time, we try to access the property, we need to check the property existence, accessing to `secureElementFlags`, then access flag value. Flatting all the value in the device session state can simply this data access, but this will add 10+ property to the device session state. WDYT?

packages/device-management-kit/src/api/command/os/SecureElementFlags.ts
```diff
@@ -0,0 +1,63 @@
...
+  type DeviceGeneralState,
+  type EndorsementInformation,
+  type OnboardingStatus,
+  type WordsInformation,
+} from "@api/device/SecureElementFlags";
+
+/**
+ * Secure element flags parser class.
+ */
+export class SecureElementFlags {
```
ofreyssinet-ledger, line 11: [SHOULD] rename to `SecureElementFlagsParser`

packages/device-management-kit/src/api/command/os/SecureElementFlags.ts
```diff
@@ -0,0 +1,63 @@
...
+   * @param seFlags - The secure element flags as a number, it comes from the response of the GetOsVersionCommand.
+   */
+  constructor(private readonly seFlags: number) {}
+
+  /**
+   * Retrieves the general device state based on the secure element flags, which is the first byte of the flags.
+   * @returns {DeviceGeneralState}
+   */
+  generalDeviceState(): DeviceGeneralState {
+    const firstByte: number = (this.seFlags >> (8 * 3)) & 0xff;
```
ofreyssinet-ledger, line 23: [COULD] Extract this in a method `getFlagNthByte(n: number)`

packages/device-management-kit/src/api/command/os/SecureElementFlags.ts
```diff
@@ -0,0 +1,63 @@
...
+
+    return {
+      isPinValidated: ((firstByte >> 7) & 1) === 1,
```
ofreyssinet-ledger, line 26: [COULD] Extract this in a function `checkByteNthBit(n: number): boolean`

packages/device-management-kit/src/api/command/os/SecureElementFlags.ts
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Secure element flags parser class.
+ */
+export class SecureElementFlags {
+  /**
+   * Creates an instance of SecureElementFlags parser.
+   * @param seFlags - The secure element flags as a number, it comes from the response of the GetOsVersionCommand.
+   */
+  constructor(private readonly seFlags: number) {}
```
ofreyssinet-ledger, line 16: [COULD] Maybe it would be easier if we pass seFlags as a Uint8Array

packages/device-management-kit/src/api/command/os/GetOsVersionCommand.ts
```diff
@@ -100,6 +118,10 @@ export class GetOsVersionCommand implements Command<GetOsVersionResponse> {
       parser.encodeToHexaString(parser.extractFieldLVEncoded()).toString(),
       16,
     );
+    const seFlagsParser = new SecureElementFlags(seFlags);
+    // This is the parsed secure element flags.
+    const secureElementFlags = { ...seFlagsParser.generalDeviceState() };
+
     const mcuSephVersion = parser.encodeToString(
```
ofreyssinet-ledger, line 125: 👍

packages/device-management-kit/src/api/device-session/DeviceSessionState.ts
```diff
@@ -68,7 +68,7 @@ type DeviceSessionBaseState = {
   readonly deviceName?: string;
 };
 
-type DeviceSessionReadyState = {
+export type DeviceSessionReadyState = {
```
ofreyssinet-ledger, line 71: [SHOULD] remove export

packages/device-management-kit/src/api/command/os/SecureElementFlags.ts
jiyuzhuang: This only change the way we get the byte, not obvious improvement.

packages/device-management-kit/src/api/command/os/SecureElementFlags.ts
```diff
@@ -0,0 +1,82 @@
...
+    return ((byte >> (8 - n)) & 1) === 1;
+  }
+
+  /**
+   * Get the nth byte of the secure element flags.
+   * @param n - The byte number to retrieve.
+   * @returns {number} The nth byte of the secure element flags.
+   */
+  _getNthByte(n: number): number {
+    return (this.seFlags >> (8 * (3 - n + 1))) & 0xff;
```
ofreyssinet-ledger, line 80: so (4 - n) no? 😁

packages/device-management-kit/src/api/command/os/SecureElementFlags.ts
ofreyssinet-ledger: [SHOULD] Unit test

ofreyssinet-ledger (changes requested):
Almost all good!
Just please:
- rename the parser to `SecureElementFlagsParser`
- unit test it

packages/device-management-kit/src/api/command/os/SecureElementFlags.ts
jiyuzhuang: I removed it as now we pass an `Uint8Array` to the parser.

ofreyssinet-ledger (approved):

PR #861, opened 2025-05-22
Feat/live 19120 improve logs

aussedatlo:
### 📝 Description

Improve logs in DMK

### ❓ Context

- **JIRA or GitHub link**:

- **Feature**:

### ✅ Checklist

  - _list of the changes_

---

### DIFF
--- a/packages/device-management-kit/src/api/logger-subscriber/model/LogLevel.ts
+++ b/packages/device-management-kit/src/api/logger-subscriber/model/LogLevel.ts
@@ -4,4 +4,5 @@ export enum LogLevel {
   Warning = 2,
   Info = 3,
   Debug = 4,
+  Verbose = 5,
 }
--- a/packages/device-management-kit/src/api/logger-subscriber/service/ConsoleLogger.ts
+++ b/packages/device-management-kit/src/api/logger-subscriber/service/ConsoleLogger.ts
@@ -1,6 +1,13 @@
 import { LogLevel } from "@api/logger-subscriber/model/LogLevel";
-import { type LogSubscriberOptions } from "@api/logger-subscriber/model/LogSubscriberOptions";
+import {
+  type LogSubscriberData,
+  type LogSubscriberOptions,
+} from "@api/logger-subscriber/model/LogSubscriberOptions";
 import { type LoggerSubscriberService } from "@api/logger-subscriber/service/LoggerSubscriberService";
+import { bufferToHexaString } from "@api/utils/HexaString";
+
+const isObject = (data: unknown): data is object =>
+  typeof data === "object" && data !== null && !Array.isArray(data);
 
 export class ConsoleLogger implements LoggerSubscriberService {
   private readonly maxLevel: LogLevel;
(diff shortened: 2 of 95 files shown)


packages/device-management-kit/src/internal/device-session/model/DeviceSession.ts
jdabbech-ledger, line 1: [ASK] I find that a logger tag with an exact class name is clearer for retrieving the corresponding logs.
If it's ok could you replace `device-session` by `DeviceSession` in constructor pls?

jdabbech-ledger (approved):
LGTM

packages/device-management-kit/src/api/logger-subscriber/model/LogLevel.ts
```diff
@@ -4,4 +4,5 @@ export enum LogLevel {
   Warning = 2,
   Info = 3,
   Debug = 4,
+  Verbose = 5,
```
valpinkman, line 7: I feel like `Debug` and `Verbose` achieve the same thing. Why do we need an extra level of logging ?

packages/device-management-kit/src/api/logger-subscriber/service/ConsoleLogger.ts
```diff
@@ -9,46 +16,73 @@ export class ConsoleLogger implements LoggerSubscriberService {
...
       case LogLevel.Debug: {
         if (this.maxLevel >= LogLevel.Debug) {
-          console.debug(tag, message, options.data);
+          console.debug(...args);
+        }
+        break;
+      }
+      case LogLevel.Verbose: {
+        if (this.maxLevel >= LogLevel.Verbose) {
+          console.debug(...args);
```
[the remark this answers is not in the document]
valpinkman: I feel like this is relevant if we keep the new `verbose` level

packages/device-management-kit/src/api/logger-subscriber/model/LogLevel.ts
aussedatlo: This is in case we need even more logs, but don't want to spam the debug as in a majority of cases debug will be sufficient

valpinkman (commented):
[ASK] Why did we introduce a new level of verbosity, when i believe `debug` === `verbose`

I'm not sure we need this level, on top of the fact that we log "Verbose" with `console.debug`

Maybe it might be interesting to change the method used to log (`console.trace` maybe ?)

Also, on the topic of improving the logger, I would add `chalk` and use it to colour our messages in the console logger depending on the verbosity (my 2 cents)

packages/device-management-kit/src/api/logger-subscriber/model/LogLevel.ts
valpinkman: Then I fill we are not using our levels correctly. I would say you should leverage `info` more and use `debug` for very specific low level stuff... idk.... 
I'm not against it, but I think it's a bit worthless

aussedatlo:
I agree with @valpinkman, we need more work on this to see what do we really need as info or debug and yes maybe verbose is not the right solution. We need to think about it.

PR #889, opened 2025-06-09
✨ (network) [DSDK-694]: Dynamic network

tom-tochito:
### 📝 Description
This PR handles dynamic networks where the network configuration command and task is created. Using the context module, the network configuration loader is created and the task is added in the eth signer action

### ❓ Context

- **JIRA or GitHub link**: [DSDK-694](https://ledgerhq.atlassian.net/browse/DSDK-694)

### ✅ Checklist

  - _list of the changes_

---

[DSDK-694]: https://ledgerhq.atlassian.net/browse/DSDK-694?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/signer/context-module/src/network/domain/NetworkConfigurationLoader.ts
```diff
@@ -0,0 +1,26 @@
...
+
+export type NetworkIcon = {
+  flex: string;
+  stax: string;
+};
+
+export type NetworkConfiguration = {
+  id: string;
+  descriptors: Record<DeviceModelId, NetworkDescriptor>;
+  icons: NetworkIcon;
```
paoun-ledger, line 21: [SHOULD] that's not a good idea to hardcode `flex` and `stax` here since we will soon have other kind of devices... that should rather be something like `icons: Record<DeviceModelId, string>;`

or it may be even better to put the optional icon in the descriptor itself, then here you would only have descriptors:
`descriptors: Record<DeviceModelId, NetworkDescriptor>;`

which would look like that:

    export type NetworkDescriptor = {
      descriptorType: string;
      descriptorVersion: string;
      data: string;
      signatures: {
        prod: string;
        test: string;
      };
      icon: string | undefined;
    };

packages/signer/context-module/src/network/data/HttpNetworkDataSource.ts
```diff
@@ -0,0 +1,91 @@
...
+      }, {} as Record<string, NetworkConfiguration["descriptors"][string]>);
+
+      const configuration: NetworkConfiguration = {
+        id: networkData.id,
+        descriptors,
+        icons: networkData.icons,
+      };
+
+      return Right(configuration);
+    } catch (error) {
```
paoun-ledger, line 83: [COULD] maybe the `try` could have a smaller scope, for instance only around axios calls?

packages/signer/signer-eth/src/internal/app-binder/command/ProvideNetworkConfigurationCommand.ts
```diff
@@ -0,0 +1,60 @@
...
+    EthAppCommandErrorFactory,
+  );
+
+  constructor(private readonly args: ProvideNetworkConfigurationCommandArgs) {}
+
+  getApdu(): Apdu {
+    const provideNetworkConfigurationArgs: ApduBuilderArgs = {
+      cla: 0xe0,
+      ins: 0x30,
+      p1: this.args.isFirstChunk ? 0x00 : 0x01,
```
paoun-ledger, line 46: [SHOULD] looking at the app documentation: https://github.com/LedgerHQ/app-ethereum/blob/develop/doc/ethapp.adoc#provide-network-configuration
it looks like you didn't put the correct values for `p1`.
Also, `p2` should be configurable deepending on the command (`configuration` vs `icon`)

packages/signer/context-module/src/network/data/HttpNetworkDataSource.ts
```diff
@@ -0,0 +1,91 @@
+import { type AxiosInstance } from "axios";
```
fAnselmi-Ledger, line 1: [SHOULD] tests are missing for this file

packages/signer/context-module/src/network/data/HttpNetworkDataSource.ts
```diff
@@ -0,0 +1,91 @@
...
+      const response = await this._api.get<NetworkApiResponse>(
+        `/v1/networks?output=id,descriptors,icons&id=${networkId}`,
+      );
+
+      const networkData = response.data.data?.[0];
+      if (!networkData) {
+        return Left(new Error(`Network configuration not found for chain ID: ${chainId}`));
+      }
+
+      const descriptors = networkData.descriptors.reduce((acc, desc) => {
```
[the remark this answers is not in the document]
paoun-ledger: Not sure about this, I think other data sources transform the backend DTO into an internal type of the ContextModule. This allows to only update the data source if the backend modify its DTO without changing its business logic. The domain class can do some higher level data transormation such as getting according certificate from another loader, and adding it into the external type. wdyt?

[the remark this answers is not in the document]
fAnselmi-Ledger: @paoun-ledger ok I see, I'd move at least transformations in its own handler tho

aussedatlo (changes requested):

aussedatlo:
- `DefaultNetworkConfigurationLoader`:
  - should implement `ContextLoader` interface, like `TokenContextLoader`
  - `load` return an array of `ClearSignContext`
- `ClearSignContext`:
  - Add a new type in `ClearSignContextType` to handle dynamic network context
  - Add the `icon`
- `Provide...Task`: edit all the provides to add the 2 dynamic network command calls

Also, NetworkConfiguration is maybe too generic, maybe rename it to something more specific like EVMNetwork or DynamicNetwork? wdyt?

aussedatlo (changes requested):
I think you have one layer not needed. You should have only one loader and one datasource

PR #996, opened 2025-08-29
📝 (keyring) [NO-ISSUE]: Decrypt binary data in the sample app

thesan:
### 📝 Description

I added this for the demo when I realized there was no way to decipher account data from the sample app:

image

### ❓ Context

- **JIRA or GitHub link**:

- **Feature**:

### ✅ Checklist

  - _list of the changes_

---


thesan:
> The PR looks okay but I'm wondering about copilot comment. Is it a risk we should look into ?

Thanks for the review.

The copilot comment looks wrong to me. I don't know how `fromCharCode` is implemented but I doubt spreading a Uint8Array would lead to a "maximum call stack size limit" on a native function.

Also `btoa(new TextDecoder().decode(bytes));` is wrong. Because `TextDecoder().decode(bytes)` will interprete the bytes as UTF-8 chars so it will return a string which might have non ASCII chars which will crash the `btoa` call:

image.

This is why this function is needed. Although with valid utf8 (like in this example ☝️) you can easily retrieve the binary representation. But if you encode random bytes in utf8 string I don't think you always keep the information, like I don't think you can go from something like `x�՗[o�H\u0014ǿK����/�a�P� �` to the correct bytes it was encoded from. This is why I think the existing UTF-8 decrypt function can't decrypt non UTF-8 strings no matter what is done with the output.

PR #1026, opened 2025-09-15
🩹 (sample) [NO-ISSUE]: Unsign ethereum tx before calling signTx

aussedatlo:
### 📝 Description

when using a raw transaction, we need to unsign it in the sample app before calling signTransaction, or transaction checks will fail as the transaction hash will be different

### ❓ Context

- **JIRA or GitHub link**:

- **Feature**:

### ✅ Checklist

  - _list of the changes_

---

### DIFF
--- a/apps/sample/src/components/SignerEthView/index.tsx
+++ b/apps/sample/src/components/SignerEthView/index.tsx
@@ -115,21 +115,22 @@ export const SignerEthView: React.FC<{ sessionId: string }> = ({
             throw new Error("Signer not initialized");
           }
 
-          // If it's a json transaction, try to parse it
+          let rawTx: string;
+
+          // Try to parse as JSON, otherwise treat as raw serialized transaction
           try {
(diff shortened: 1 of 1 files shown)


smartine-ledger (commented):
Sorry. I just removed my approval because I was wondering myself : IMHO this check should be in the signer part and not in the sample part. WDYT?

aussedatlo:
> > We discussed this, and there’s no real-world scenario where the signTransaction function would be called with an unsigned transaction. Therefore, it’s better to let the sample app handle this edge case, which is only used to replay transaction signatures with already signed transactions.
> 
> Ok then but I don't understand : if it's not a real world scenario why do we want to handle this case?

Because it's easier for us to replay signed transactions from etherscan to debug instead of crafting unsigned transactions, but in real life you don't want to sign a transaction that is already signed

ofreyssinet-ledger (approved):

PR #1031, opened 2025-09-16
🐛 (rn-ble): Fix discovery of connected devices & reconnection failure…

ofreyssinet-ledger:
… cleanup

### 📝 Description

There is currently an issue on the OS update in Ledger Live that is due to some stuff in the React Native BLE transport.

It happens at the end of the OS update: the device gets disconnected a final time to reboot, and it takes ~10/15s, longer than our reconnection timeout.

There is some kind of race condition between  the disconnection timeout inside the `DeviceConnectionStateMachine`, which triggers a `safeCancel` on the device, but the last reconnection try is still running at that point. My interpretation is that at that point, we end up in a bastard state where
 - We don't have a `DeviceConnection` anymore regarding that device because of the timeout.
 - The last reconnection trial results in a device that is partially connected, with some missing cleanup, so the device won't appear in subsequent scans, as it's "connected" (but not registered internally as a `DeviceConnection` in our transport)

The logic in LL when we lose the connection is to restart a discovery and wait for a device with the same ID to appear. But in that scenario, the device was never appearing, so it was just timing out.

**The fixes:**

I did 2 things that would each independently solve the issue, but having both will make the transport more robust overall:

- In `listenToAvailableDevices`, we merge "internally connected devices" with "scanned devices". Instead of using "internally connected devices", we now get the list of connected devices directly from the `BleManager`.
- In `tryToReconnect`, we always call `safeCancel` to clean up existing connection to the device _before_ all the retries, and _after each failure,_. It will prevent the bastard state where we didn't manage to reconnect but somehow the device is still in a pseudo connected state.

### ❓ Context

- **JIRA or GitHub link**:

- **Feature**:

### ✅ Checklist

  - _list of the changes_

---


packages/transport/rn-ble/src/api/transport/RNBleTransport.ts
```diff
@@ -259,6 +247,46 @@ export class RNBleTransport implements Transport {
...
+
+        const connectedDevices = Either.rights(eitherConnectedDevices);
+
+        this._logger.debug("[listenToAvailableDevices]", {
+          data: {
+            rawConnectedDevices: rawConnectedDevices.map((d) => d.name),
+            connectedDevices: connectedDevices.map((d) => d.name),
+            scannedDevices: scannedDevices.map((d) => d.name),
+          },
+        });
```
ofreyssinet-ledger, line 287: This logging will be useful in case we encounter scanning issues.

packages/transport/rn-ble/src/api/transport/RNBleTransport.test.ts
ofreyssinet-ledger, line 1: Refacto largely done by cursor because it was _nasty_ previously.
I also had to rework some of the mocking to be compatible with how listenToAvailableDevices works now.

aussedatlo (approved):

PR #1336, opened 2026-03-04
✨ (signer-eth) [NO-ISSUE]: Add device action factories for GetAddress and SignPersonalMessage

pdeville-ledger:
### 📝 Description

Add standalone factory functions for `GetAddress` and `SignPersonalMessage` device actions in the Ethereum signer package. These factories create `CallTaskInAppDeviceAction` instances, making it easier to instantiate device actions without going through the `EthAppBinder`.

Changes:
- New `GetAddressDeviceActionFactory` with support for `checkOnDevice`, `returnChainCode`, `skipOpenApp`, and optional `chainId`
- New `SignPersonalMessageDeviceActionFactory` with support for string and `Uint8Array` messages
- Unit tests for both factories
- Export `SignPersonalMessageDARequiredInteraction` type (previously unexported)
- Export both factories from the public API index

### ❓ Context

- **JIRA or GitHub link**: NO-ISSUE

### ✅ Checklist

  - New public API exports: `GetAddressDeviceActionFactory`, `SignPersonalMessageDeviceActionFactory`
  - Exported `SignPersonalMessageDARequiredInteraction` type

### DIFF
--- a/packages/signer/signer-eth/src/api/app-binder/GetAddressDeviceActionFactory.ts
+++ b/packages/signer/signer-eth/src/api/app-binder/GetAddressDeviceActionFactory.ts
@@ -1,0 +1,48 @@
+import { type ContextModule } from "@ledgerhq/context-module";
+import {
+  CallTaskInAppDeviceAction,
+  type LoggerPublisherService,
+  UserInteractionRequired,
+} from "@ledgerhq/device-management-kit";
+
+import { type GetAddressCommandResponse } from "@api/app-binder/GetAddressCommandTypes";
+import { type EthErrorCodes } from "@internal/app-binder/command/utils/ethAppErrors";
+import { APP_NAME } from "@internal/app-binder/constants";
+import { SendGetAddressTask } from "@internal/app-binder/task/SendGetAddressTask";
+
+export const GetAddressDeviceActionFactory = (args: {
+  derivationPath: string;
+  checkOnDevice: boolean;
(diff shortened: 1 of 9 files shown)


packages/signer/signer-eth/src/api/app-binder/GetAddressDeviceActionFactory.ts
```diff
@@ -0,0 +1,47 @@
...
+      task: async (internalApi) =>
+        new SendGetAddressTask(internalApi, {
+          contextModule: args.contextModule,
+          derivationPath: args.derivationPath,
+          checkOnDevice: args.checkOnDevice,
+          returnChainCode: args.returnChainCode,
+          chainId: args.chainId,
+          loggerFactory: args.loggerFactory,
+        }).run(),
+      appName: "Ethereum",
```
smartine-ledger: [COMMENT] What about creating and exposing  a "const string appName = Ethereum". This way you can use the same field everywhere (DRY principle)

smartine-ledger (approved):
Nice implementation to avoid using all the complexity of the EthBinder 👍 . Just a few comment

aussedatlo:
Hey, you are introducing a new concept, DeviceActionFactory. Actually you could just use classic Device action class to do this. Do we really want to mix Factory and simple class on the codebase? If factory is the way to go, we should put all DA in internal and only put factories in api.

PR #1377, opened 2026-03-22
fix(node-hid): filter Ledger devices to the APDU interface

danielnordh:
## Summary
- filter Node HID Ledger devices to the APDU HID interface on macOS and Windows
- apply the same filter during both device discovery paths in `NodeHidTransport`
- add regression tests covering duplicate Ledger HID interfaces on macOS and Windows

## Why
On macOS and Windows, some Ledger devices expose multiple HID interfaces. Legacy LedgerJS filters these devices to the APDU interface (`usagePage === 0xffa0`) before transport selection, but `@ledgerhq/device-transport-kit-node-hid` currently filters only by `vendorId`.

In local testing, that caused `SignerEthBuilder(...).getAddress(...)` to stall on a Ledger Nano X over Node HID with firmware `2.6.1` and Ethereum app `1.21.3`. Applying the same APDU-interface filter restored correct behavior.

Other device families already worked before this filter and continued to work after it, which suggests the underlying Node HID interface-selection behavior differs across device families and/or firmware/app-version combinations.

## Validation
- verified locally against the affected Nano X combination above via DMK Node HID
- verified locally against other device families via DMK Node HID
- added unit tests for ignoring the non-APDU Ledger interface on `darwin` and `win32`

## Notes
- the currently published stable `@ledgerhq/device-transport-kit-node-hid@1.0.0` still advertises `@ledgerhq/device-management-kit: 1.1.0` as a peer, while the rest of the published signer stack is already on the `^1.2.0` line; the current repo source appears broader/aligned, so a new stable publish from current source would also help downstream consumers avoid `--legacy-peer-deps`
- I was not able to run the upstream package test suite in this environment because the clone does not have its pnpm toolchain installed locally, but the new tests are included in this patch.

### DIFF
--- a/packages/transport/node-hid/src/api/transport/NodeHidTransport.test.ts
+++ b/packages/transport/node-hid/src/api/transport/NodeHidTransport.test.ts
@@ -368,6 +368,37 @@ describe("NodeHidTransport", () => {
       ).toBe("/dev/hidraw1");
     });
 
+    it("should ignore non-APDU ledger interfaces on win32", async () => {
+      setProcessPlatform("win32");
+
+      const genericInterface = nodeHidDeviceStubBuilder({
+        path: "\\\\?\\hid#vid_2c97&pid_0001&mi_01#generic",
+        interface: 2,
+        usagePage: 0xf1d0,
+      });
+      const apduInterface = nodeHidDeviceStubBuilder({
+        path: "\\\\?\\hid#vid_2c97&pid_0001&mi_00#apdu",
+        interface: 0,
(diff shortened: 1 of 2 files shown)


OlivierFreyssinet (approved):
Thanks for the fix!

OlivierFreyssinet:
Thanks for this fix @danielnordh! I've recreated it as #1399 with signed commits, formatted commit messages, and a changeset to pass our CI checks. You're added as co-author on the commit.

Closing this one in favour of #1399.

PR #1384, opened 2026-03-24
🎨 (clear-sign-tester): Add --external-speculos flag to skip Docker lifecycle

Maroutis:
## Summary

- Add `--external-speculos` CLI option to `cs-tester` that skips the internal Docker Speculos lifecycle and connects to an already-running Speculos instance.
- When the flag is set, `SpeculosServiceController` is excluded from the service controllers array so no Docker container is started/stopped.
- In external mode, ports default to Speculos' standard values (5000 API / 8000 VNC) instead of random host ports; override with `--speculos-port` / `--speculos-vnc-port` if needed.
- Enables deployments where Speculos runs natively (e.g. CI runners, AWS App Runner containers with `qemu-user-static`) to avoid Docker-in-Docker overhead.

## Context

We are deploying [erc7730-analyzer](https://github.com/LedgerHQ/erc7730-analyzer) on a server that cannot run Docker. The cs-tester currently always spawns Docker-based Speculos instances via `SpeculosServiceController`, even when a Speculos instance is already provided via `--speculos-port`. This PR adds `--external-speculos` to bypass that behavior entirely.

## Changes

| File | Change |
|------|--------|
| `SpeculosConfig.ts` | New `externalSpeculos?: boolean` field |
| `EthereumTransactionTesterCli.ts` | New `--external-speculos` CLI option; sensible port defaults in external mode |
| `infrastructureModuleFactory.ts` | Conditionally skip `SpeculosServiceController` when flag is set |

## Test plan

- [ ] Run `cs-tester` **without** `--external-speculos` → Docker Speculos lifecycle unchanged (no regression)
- [ ] Start Speculos natively, run `cs-tester --external-speculos --speculos-port 5555` → connects to external instance, produces screenshots
- [ ] Run `cs-tester --external-speculos` without explicit port → defaults to 5000
- [ ] Verify `--help` output includes the new flag with updated documentation

### DIFF
--- a/apps/clear-signing-tester/src/cli/EthereumTransactionTesterCli.ts
+++ b/apps/clear-signing-tester/src/cli/EthereumTransactionTesterCli.ts
@@ -36,6 +36,7 @@ export type CliConfig = {
   screenshotFolderPath?: string;
   customApp?: string;
   forcePull?: boolean;
+  externalSpeculos?: boolean;
 
   // config.signer
   skipCal?: boolean;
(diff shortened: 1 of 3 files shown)


Maroutis:
Closing — recreating with correct branch naming convention.

PR #1392, opened 2026-03-27
🎨 (context-module) [NO-ISSUE]: Refactor ContextModuleConfig to restrict public API surface

aussedatlo:
### 📝 Description

Refactor `ContextModuleConfig` into separate, focused types to improve separation of concerns and restrict the public API surface:

- **`ContextModuleConfig`**: Only contains service URL configurations (cal, web3checks, metadataServiceDomain, reporter, datasource) — this is the type exposed publicly.
- **`ContextModuleServiceConfig`**: Extends `ContextModuleConfig` with runtime dependencies (`originToken`, `loggerFactory`) — internal only.
- **`ContextModuleLoaderConfig`**: Contains loader-related configuration (custom loaders, field loaders, etc.) — internal only.

Additional improvements:
- Default config objects are now frozen with `Object.freeze()` to prevent accidental mutations.
- Module factories (`proxyModuleFactory`, `trustedNameModuleFactory`) now receive specific config slices instead of the full config object.
- Public exports in `index.ts` are now explicitly listed to avoid leaking internal types.
- All HTTP data sources now depend on `ContextModuleServiceConfig` instead of the monolithic `ContextModuleConfig`.

### ❓ Context

### ✅ Checklist

  - Internal types (`ContextModuleLoaderConfig`) are no longer exported publicly
  - `ContextModuleConfig` type is narrower (only service URLs), which is a breaking type change for consumers who relied on the full type
  - Default config is now immutable (frozen)
  - All data sources and module factories use more precise config types

### DIFF
--- a/packages/signer/context-module/src/config/model/ContextModuleConfig.ts
+++ b/packages/signer/context-module/src/config/model/ContextModuleConfig.ts
@@ -36,16 +36,22 @@ export type ContextModuleConfig = {
   cal: ContextModuleCalConfig;
   web3checks: ContextModuleWeb3ChecksConfig;
   metadataServiceDomain: ContextModuleMetadataServiceConfig;
+  reporter: ContextModuleReporterConfig;
+  datasource: ContextModuleDatasourceConfig;
+};
+
+export type ContextModuleServiceConfig = ContextModuleConfig & {
+  originToken: string;
+  loggerFactory: (tag: string) => LoggerPublisherService;
+};
+
+export type ContextModuleLoaderConfig = {
   defaultLoaders: boolean;
   defaultFieldLoaders: boolean;
   customFieldLoaders: ContextFieldLoader[];
   customLoaders: ContextLoader[];
-  loggerFactory: (tag: string) => LoggerPublisherService;
   customTypedDataLoader?: TypedDataContextLoader;
   customSolanaLoader?: SolanaContextLoader;
   customBlindSigningReporter?: BlindSigningReporter;
   customTrustedNameDataSource?: TrustedNameDataSource;
-  originToken?: string;
-  datasource?: ContextModuleDatasourceConfig;
-  reporter?: ContextModuleReporterConfig;
 };
--- a/.changeset/lovely-aliens-report.md
+++ b/.changeset/lovely-aliens-report.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/context-module": minor
+---
+
+Refactor ContextModuleConfig to only expose necessary configuration types publicly, keeping internal types like loader config private since they are not intended for external consumers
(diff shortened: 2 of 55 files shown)


fAnselmi-Ledger (approved):

packages/signer/context-module/src/config/model/ContextModuleConfig.ts
```diff
@@ -36,16 +36,22 @@ export type ContextModuleConfig = {
   cal: ContextModuleCalConfig;
   web3checks: ContextModuleWeb3ChecksConfig;
   metadataServiceDomain: ContextModuleMetadataServiceConfig;
+  reporter: ContextModuleReporterConfig;
+  datasource: ContextModuleDatasourceConfig;
+};
+
+export type ContextModuleServiceConfig = ContextModuleConfig & {
```
smartine-ledger, line 43: [ASK]  You are creating a strong dependency here. Why `ContextModuleServiceConfig` needs to extend `ContextModuleConfig`?

smartine-ledger (approved):

packages/signer/context-module/src/config/model/ContextModuleConfig.ts
aussedatlo: ContextModuleServiceConfig contains url and other stuff, so it's basically ContextModuleConfig plus some attribute like logger. it's to avoid duplicating attribute

smartine-ledger: Hum... Ok. 100% personal but I prefer composing config than using Inheritance. I mean :
```
type ContextModuleServiceConfig {
  ....
  moduleConfig: ContextModuleConfig
}
```
It' up to you 😉

smartine-ledger: In addtition : a service in a config... Seems weird IMHO. 🤷

smartine-ledger (approved):

packages/signer/context-module/src/config/model/ContextModuleConfig.ts
aussedatlo: ContextModuleConfig is already used by externals but contain too much attribute, like private stuff. ContextModuleServiceConfig is the config for services like datasources that will use urls loggers and so on. it's not really a service in a config but a config used by some services

aussedatlo: Do you have another name to suggest?

smartine-ledger: My bad. I was talking about this line in the config : `loggerFactory: (tag: string) => LoggerPublisherService;`.  😅

aussedatlo: yes you are right, this is in there only because we need to set up the DI but could be a specific type for DI args, but is was already like this before. will keep it in mind and will patch it when I will have time 👍

PR #1411, opened 2026-04-08
✨ (os-update) [DSDK-1043]: Add backup Device Action

benruseau:
### 📝 Description

Implement the `BackupDeviceAction` as part of the coming OS update migration to the DMK.

This device action orchestrates a full device backup before an OS update. It collects the device language, installed apps storage data, and custom lock screen image, then persists them in injected `KeyValueStorage`.

💡  You can test the device action in the `Sample` app, where `KeyValueStorage` has been adapted to store the backup inside the web browser local storage (see the demo video just below).

**State machine flow:**

1. **LookForBackup** — checks if a fresh backup (< 24h) already exists for the device; if so, short-circuits to `Success`
2. **GetLanguage** — retrieves the current language ID from the device
3. **ListInstalledApps** — lists all apps installed on the device (skipped if the device is unseeded)
4. **BackupAppsStorage** — iterates over installed apps and backs up each app's storage data via `BackupAppStorageTask`
5. **DownloadCustomLockScreen** — downloads the custom lock screen image (skipped on devices that don't support it: only STAX, FLEX, and APEX)
6. **SaveBackup** — persists the backup (language, apps storage, CLS image, timestamp) in the `KeyValueStorage` wrapped in `DeviceBackupDataSource`

The action outputs nothing on success, or a typed error on failure.

**Result:**

https://github.com/user-attachments/assets/7dd7848f-bc59-427c-b533-5f800e2f83c7

### ❓ Context

- **JIRA link**: [DSDK-1043](https://ledgerhq.atlassian.net/jira/software/c/projects/DSDK/boards/817?quickFilter=2705&selectedIssue=DSDK-1043)

- **Feature**: Backup Device Action — prerequisite step for OS update migration to the DMK

### ✅ Checklist

- [x] **Covered by automatic tests** — unit tests cover all substeps and the full state machine (happy path + error paths + skip conditions)
  - New `BackupDeviceAction` class in `packages/ledger-wallet`
  - New `KeyValueStorage` interface (consumers must provide an implementation)
  - New error types: `LookForBackupError`, `GetLanguageIdError`, `BackupAppStorageError`, `SaveBackupError`...
  - New `BackupSteps` enum and `BackupData`/`BackupApp` types exported from `types.ts`
  - QA should verify: backup short-circuit when a recent backup exists, correct skip of CLS step on non-STAX/FLEX/APEX devices, skip of app listing on unseeded devices, and error propagation for each substep

---

### DIFF
--- a/packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/BackupDeviceAction.ts
+++ b/packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/BackupDeviceAction.ts
@@ -1,0 +1,428 @@
+import {
+  bufferToHexaString,
+  type DeviceActionStateMachine,
+  DeviceModelId,
+  type InternalApi,
+  type StateMachineTypes,
+  UserInteractionRequired,
+  XStateDeviceAction,
+} from "@ledgerhq/device-management-kit";
+import { Left, Right } from "purify-ts";
+import { assign, fromPromise, setup } from "xstate";
+
+import { DeviceBackupStorage } from "@api/device-action/OsUpdate/Backup/Storage/DeviceBackupStorage";
+import { backupAppsStorage } from "@api/device-action/OsUpdate/Backup/Substeps/BackupAppsStorage";
+import { downloadCustomLockScreenDevice } from "@api/device-action/OsUpdate/Backup/Substeps/DownloadCustomLockScreen";
+import { getLanguageId } from "@api/device-action/OsUpdate/Backup/Substeps/GetLanguageId";
+import { listInstalledApps } from "@api/device-action/OsUpdate/Backup/Substeps/ListInstalledApps";
+import { lookForBackup } from "@api/device-action/OsUpdate/Backup/Substeps/LookForBackup";
+import { saveBackup } from "@api/device-action/OsUpdate/Backup/Substeps/SaveBackup";
+import {
+  type BackupDAError,
+  type BackupDAInput,
+  type BackupDAIntermediateValue,
+  type BackupDAInternalState,
+  type BackupDAOutput,
+  BackupSteps,
+} from "@api/device-action/OsUpdate/Backup/types";
+
+export class BackupDeviceAction extends XStateDeviceAction<
+  BackupDAOutput,
+  BackupDAInput,
+  BackupDAError,
+  BackupDAIntermediateValue,
+  BackupDAInternalState
+> {
+  protected override makeStateMachine(
+    internalAPI: InternalApi,
+  ): DeviceActionStateMachine<
+    BackupDAOutput,
+    BackupDAInput,
+    BackupDAError,
+    BackupDAIntermediateValue,
+    BackupDAInternalState
+  > {
+    type types = StateMachineTypes<
+      BackupDAOutput,
+      BackupDAInput,
+      BackupDAError,
+      BackupDAIntermediateValue,
+      BackupDAInternalState
+    >;
+
+    const { storage } = this.input;
+    const deviceBackupStorage = new DeviceBackupStorage(storage);
+
+    return setup({
+      types: {
+        input: {} as types["input"],
+        output: {} as types["output"],
+        context: {} as types["context"],
+      } as types,
+      actors: {
+        lookForBackup: fromPromise(lookForBackup(deviceBackupStorage)),
+        getLanguageId: fromPromise(getLanguageId(internalAPI)),
+        listInstalledApps: listInstalledApps(
+          internalAPI,
+          this.input.unlockTimeout,
+        ),
+        downloadCustomLockScreen: downloadCustomLockScreenDevice(
+          internalAPI,
+          this.input.unlockTimeout,
+          true,
+        ),
+        backupAppsStorage: fromPromise(
+          backupAppsStorage(internalAPI, this.getLoggerFactory(internalAPI)),
+        ),
+        saveBackup: fromPromise(saveBackup(deviceBackupStorage)),
+      },
+      guards: {
+        hasBackup: ({ context }) =>
+          context._internalState.backupAlreadyExist === true,
+        isDeviceOnboarded: ({ context }) => context.input.isDeviceOnboarded,
+        isCustomLockScreenFeatureSupported: () =>
+          [DeviceModelId.APEX, DeviceModelId.FLEX, DeviceModelId.STAX].includes(
+            internalAPI.getDeviceModel().id,
+          ),
+        hasError: ({ context }) => context._internalState.error !== null,
+      },
+      actions: {
+        assignErrorFromEvent: assign({
+          _internalState: (_) => ({
+            ..._.context._internalState,
+            error: _.event["error"], // NOTE: should never happen
+          }),
+        }),
+      },
+    }).createMachine({
+      id: "BackupDeviceAction",
+      initial: "LookForBackup",
+      context: ({ input }) => ({
+        input: {
+          isDeviceOnboarded: input.isDeviceOnboarded,
+          deviceId: input.deviceId,
+          storage: input.storage,
+          unlockTimeout: input.unlockTimeout,
+        },
+        intermediateValue: {
+          requiredUserInteraction: UserInteractionRequired.None,
+          step: BackupSteps.Idle,
+        },
+        _internalState: {
+          error: null,
+          languageId: undefined,
+          installedApps: [],
+          backupApps: [],
+          clsHexImage: undefined,
+          backupAlreadyExist: false,
+        },
+      }),
+      states: {
+        LookForBackup: {
+          invoke: {
+            src: "lookForBackup",
+            input: ({ context }) => ({
+              deviceId: context.input.deviceId,
+            }),
+            onDone: {
+              actions: assign({
+                _internalState: (_) => {
+                  return _.event.output.caseOf<BackupDAInternalState>({
+                    Left: (error) => ({
+                      ..._.context._internalState,
+                      error,
+                    }),
+                    Right: (backupAlreadyExist) => ({
+                      ..._.context._internalState,
+                      backupAlreadyExist,
+                    }),
+                  });
+                },
+              }),
+              target: "CheckIfBackupExist",
+            },
+            onError: {
+              actions: "assignErrorFromEvent",
+              target: "Error",
+            },
+          },
+        },
+        CheckIfBackupExist: {
+          always: [
+            {
+              guard: "hasError",
+              target: "Error",
+            },
+            {
+              guard: "hasBackup",
+              target: "Success",
+            },
+            {
+              target: "GetLanguage",
+            },
+          ],
+        },
+        GetLanguage: {
+          entry: assign({
+            intermediateValue: (_) => ({
+              ..._.context.intermediateValue,
+              step: BackupSteps.GetLanguage,
+            }),
+          }),
+          invoke: {
+            src: "getLanguageId",
+            onDone: {
+              actions: assign({
+                _internalState: (_) => {
+                  return _.event.output.caseOf<BackupDAInternalState>({
+                    Left: (error) => ({
+                      ..._.context._internalState,
+                      error,
+                    }),
+                    Right: (languageId) => ({
+                      ..._.context._internalState,
+                      languageId,
+                    }),
+                  });
+                },
+              }),
+              target: "CheckIfDeviceIsOnboarded",
+            },
+            onError: {
+              actions: "assignErrorFromEvent",
+              target: "Error",
+            },
+          },
+        },
+        CheckIfDeviceIsOnboarded: {
+          always: [
+            {
+              guard: "hasError",
+              target: "Error",
+            },
+            {
+              guard: "isDeviceOnboarded",
+              target: "ListInstalledApps",
+            },
+            {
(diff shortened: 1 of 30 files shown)


.changeset/small-poets-enter.md
```diff
@@ -0,0 +1,6 @@
+---
+"@ledgerhq/dmk-ledger-wallet": minor
+"@ledgerhq/device-management-kit": patch
```
aussedatlo: [SHOULD] Split the changeset, one changeset per package 🙏 see https://github.com/LedgerHQ/device-sdk-ts/blob/develop/.cursor/rules/changeset.mdc#L21

apps/sample/src/utils/os-update/localStorageDeviceBackupDataSource.ts
```diff
@@ -0,0 +1,29 @@
+import {
+  type BackupData,
+  type DeviceBackupDataSource,
+} from "@ledgerhq/dmk-ledger-wallet/api/device-action/OsUpdate/Backup/types.js";
```
aussedatlo: [SHOULD] fix `.js` import

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/Substeps/BackupAppsStorage.ts
```diff
@@ -0,0 +1,55 @@
...
+  type LoggerPublisherService,
+} from "@ledgerhq/device-management-kit";
+import { type Either, Left, Right } from "purify-ts";
+
+import { BackupAppStorageError } from "@api/device-action/OsUpdate/Backup/BackupDeviceActionErrors";
+import { type BackupApp } from "@api/device-action/OsUpdate/Backup/types";
+
+const EMPTY_STORAGE = "0x";
+
+export const backupAppsStorage =
```
aussedatlo, line 27: [COULD] use a type for both args and response type? I find it clearer but that just my opinion

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/Substeps/BackupAppsStorage.ts
```diff
@@ -0,0 +1,55 @@
...
+    for (const installedApp of installedApps) {
+      const backupAppStorageTaskResponse = await new BackupAppStorageTask(
+        {
+          appName: installedApp.name,
+        },
+        internalApi,
+        logger,
+      ).run();
+
+      if (backupAppStorageTaskResponse.status === CommandResultStatus.Error) {
```
aussedatlo: We have an utils to do this `isSuccessCommandResult` and should help you to infer type correctly (I think)

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/Substeps/DownloadCustomLockScreen.test.ts
```diff
@@ -0,0 +1,49 @@
+import { makeDeviceActionInternalApiMock } from "@api/device-action/__test-utils__/makeInternalApi";
+import { DownloadCustomLockScreenDeviceAction } from "@api/device-action/DownloadCustomLockScreen/DownloadCustomLockScreenDeviceAction";
+
+import { downloadCustomLockScreenDevice } from "./DownloadCustomLockScreen";
+
+vi.mock(
```
aussedatlo, line 5: [ASK] What this test is doing? you are testing `DownloadCustomLockScreenDeviceAction` but also mocking it? 

note: Usually we use injection pattern to avoid using `vi.mock`

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/Substeps/GetBackupIfExist.ts
```diff
@@ -0,0 +1,37 @@
...
+
+import { GetBackupIfExistError } from "@api/device-action/OsUpdate/Backup/BackupDeviceActionErrors";
+import {
+  type BackupData,
+  type DeviceBackupDataSource,
+} from "@api/device-action/OsUpdate/Backup/types";
+
+const isBackupDataExpired = (backupData: BackupData): boolean => {
+  const backupAge = Date.now() - new Date(backupData.createdAt).getTime();
+  const oneDayInMs = 24 * 60 * 60 * 1000;
```
aussedatlo: [COULD] use const in uppercase for magic numbers: `ONE_DAY_IN_MS`

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/Substeps/GetLanguageId.ts
```diff
@@ -0,0 +1,20 @@
...
+import { type Either, Left, Right } from "purify-ts/Either";
+
+import { GetLanguageIdError } from "@api/device-action/OsUpdate/Backup/BackupDeviceActionErrors";
+
+export const getLanguageId =
+  (internalApi: InternalApi) =>
+  async (): Promise<Either<GetLanguageIdError, number | undefined>> => {
+    const result = await internalApi.sendCommand(new GetOsVersionCommand());
+
+    if (result.status === CommandResultStatus.Error) {
```
aussedatlo: isSuccessCommandResult?

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/BackupDeviceAction.ts
```diff
@@ -0,0 +1,434 @@
...
+                  return _.event.output.caseOf<BackupDAInternalState>({
+                    Left: (error) => ({
+                      ..._.context._internalState,
+                      error,
+                    }),
+                    Right: (output) => ({
+                      ..._.context._internalState,
+                      clsHexImage:
+                        "imageData" in output
+                          ? Buffer.from(output.imageData).toString("hex")
```
aussedatlo: Avoid Buffer and prefer UInt8Array in DMK. this should wotk as this package is intended for LW and buffer is working in there, but as Buffer is not available everytime, we prefer to use UInt8Array. We also have some utils to convert array in hex string 👍

aussedatlo (changes requested):
Nice job, easy to read and good split !

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/Substeps/BackupAppsStorage.ts
benruseau: I agree!

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/Substeps/DownloadCustomLockScreen.test.ts
benruseau: It just tests that the wrapper function passes the arguments and returns what the DA produces (what's returned from `makeStateMachine`). 

If the substep explicit job is to create the DA, injecting it would just move the responsibility up one level, which defeats the point of having the substep 🤔

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/types.ts
```diff
@@ -0,0 +1,85 @@
...
+  installedApps: BackupApp[];
+  clsHexImage: string | undefined;
+  createdAt: Date;
+};
+
+export interface DeviceBackupDataSource {
+  getDataForDevice(backupKey: string): Promise<BackupData | null>;
+  saveDataForDevice(backupKey: string, data: BackupData): Promise<void>;
+  removeDataForDevice(backupKey: string): Promise<void>;
+}
```
OlivierFreyssinet: [ASK] Wouldn't it be better if the arg of those methods was instead `deviceId` ?
And the concern of how to index the data (i.e generating `backupKey`) is implemented in the data source itself?
That way, consumers cannot arbitrarily decide how data for a device is stored, or override it. They just pass the device ID and the data source decides how to store/retrieve/delete it.

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/BackupDeviceAction.ts
```diff
@@ -0,0 +1,435 @@
...
+              allowedEmpty: true,
+            }),
+            onSnapshot: {
+              actions: assign({
+                intermediateValue: (_) => ({
+                  ..._.context.intermediateValue,
+                  requiredUserInteraction:
+                    _.event.snapshot.context.intermediateValue
+                      .requiredUserInteraction,
+                }),
```
OlivierFreyssinet, line 329: [SHOULD] use the entire intermediate value of the invoked device action, not just the required user interaction. For instance with this one it's important because it also emits a `progress` value which can be quite useful if the consuming app wants to display it.

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/BackupDeviceAction.ts
```diff
@@ -0,0 +1,435 @@
...
+        Error: {
+          type: "final",
+        },
+      },
+      output: ({ context }) => {
+        if (context._internalState.error !== null) {
+          return Left(context._internalState.error);
+        }
+        return Right({
+          backupKey: context._internalState.backupKey!,
```
OlivierFreyssinet: I think it will not be required to return the backupKey in case the data source handles it internally.

OlivierFreyssinet (changes requested):
Nice job overall 👏 
Just make sure to use the intermediate value of the CLS device actions (required change), and there's potentially something to rethink about the storage (up to debate).

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/Storage/DeviceBackupStorage.ts
```diff
@@ -0,0 +1,53 @@
...
+
+  async getBackup(
+    deviceId: DeviceId,
+  ): Promise<Either<GetBackupError, Backup | null>> {
+    return EitherAsync(async () => {
+      const data = await this.storage.getItem(
+        `${BACKUP_STORAGE_PREFIX_KEY}-${deviceId}`,
+      );
+      if (!data) return null;
+      const backupData = JSON.parse(data) as BackupData;
```
OlivierFreyssinet: [COULD] be good to avoid blind type assertion like this and implement a simple type predicate to verify that the parsed data has the right format.
This would be useful in the future in case the data structure changes without a migration:
if `BackupData` changes from a version 1 to a version 2 and some users have a persisted data in version 1 (with a missing field compared to version 2 for instance), it could cause bugs.

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/Storage/DeviceBackupStorage.test.ts
OlivierFreyssinet, line 1: [SHOULD] Could we add a round-trip test here? Right now `saveBackup` and `getBackup` are tested independently, but not that the payload produced by `saveBackup` can be read back by `getBackup` into the same `Backup`. Concretely, you could call `saveBackup`, capture the string passed to `setItem`, then mock `getItem` to return that same string and assert the result is `toStrictEqual` to the original input. This would validate the `createdAt` conversion and maybe other JSON parsing subtleties.

This would be even easier to test if serialising and deserializing were done in a separate function. Then you could just test `expect(myBackup).toEqual(deserialize(serialize(myBackup))`. So either test the round trip or extract the serialization logic and test it separately.

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/Storage/DeviceBackupStorage.ts
benruseau: - Deported serialization and deserialization logic to dedicated util methods in `BackupSerializer`. 
- Added backup data validation using `purify-ts` `Codec` util

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/Storage/DeviceBackupStorage.test.ts
benruseau: Done!

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/Storage/BackupSerializer.ts
```diff
@@ -0,0 +1,56 @@
...
+  languageId: optional(number),
+  installedApps: array(
+    Codec.interface({
+      appName: string,
+      data: optional(string),
+    }),
+  ),
+  clsHexImage: optional(string),
+  createdAt: string,
+});
```
OlivierFreyssinet, line 30: Nice new usage of Purify, I didn't know this feature was part of it. It would be worth presenting it in a continuous improvement 👍

OlivierFreyssinet (approved):
great 🚀

packages/ledger-wallet/src/api/device-action/OsUpdate/Backup/Storage/BackupSerializer.ts
benruseau: good idea!

aussedatlo (approved):

PR #1446, opened 2026-04-24
♻️ (dmk) [NO-ISSUE]: Extract a generic result base DmkResult from CommandResult

### LINKED ISSUE #1442: Tech debt: stop using CommandResult outside commands
## Summary

`CommandResult` and `CommandResultFactory` are command-layer concepts, but some task and workflow abstractions currently use them as a generic result type.

That leaks task-specific errors into the central command result surface. In practice, this is why `InvalidGetFirmwareMetadataResponseError` still appears in `CommandResult` and `CommandResultFactory`: removing it cleanly is blocked by non-command code returning command-shaped results.

For example, [`CommandResultFactory` in `CommandResult.ts`](https://github.com/LedgerHQ/device-sdk-ts/blob/develop/packages/device-management-kit/src/api/command/model/CommandResult.ts) currently accepts a task-specific error in a generic command factory:

```ts
export function CommandResultFactory<Data, SpecificErrorCodes = void>({
  data,
  error,
}:
  | { data: Data; error?: undefined }
  | {
      data?: undefined;
      error:
        | DeviceExchangeError<SpecificErrorCodes>
        | InvalidResponseFormatError
        | InvalidStatusWordError
        | InvalidGetFirmwareMetadataResponseError
        | UnknownDeviceExchangeError;
    }): CommandResult<Data, SpecificErrorCodes> {
```

This is also the case for battery-command-related errors, which are being cleaned up separately in [#1441](https://github.com/LedgerHQ/device-sdk-ts/pull/1441).

## Current offenders

The clearest current non-command usages are:

- `packages/device-management-kit/src/api/device-action/task/GetFirmwareMetadataTask.ts`
- `packages/device-management-kit/src/api/device-action/task/GetApplicationsMetadataTask.ts`
- `packages/device-management-kit/src/api/device-action/task/BackupAppStorageTask.ts`
- `packages/device-management-kit/src/api/device-action/os/CallTaskInAppDeviceAction/CallTaskInAppDeviceActionTypes.ts`

These are the main reason task-specific errors such as `InvalidGetFirmwareMetadataResponseError` currently pollute `packages/device-management-kit/src/api/command/model/CommandResult.ts`.

## Why this is a problem

- `CommandResult` becomes a catch-all result type instead of a command-specific one.
- Task-specific errors are forced into command-layer unions.
- `CommandResultFactory` accumulates orchestration concerns that should not belong to APDU command parsing.
- It becomes harder to simplify command error types without touching higher-level workflows.

## Candidate solutions

1. Introduce a more generic `Result` or `DmkResult` type for tasks and other non-command workflows.
2. Migrate task wrappers such as `CallTaskInAppDeviceAction` away from `CommandResult`.
3. Consider a lint rule that restricts `CommandResult` and `CommandResultFactory` usage to the command layer, with a small allowlist where true pass-through is still required.

## Expected outcome

- Commands keep returning `CommandResult`.
- Tasks and higher-level orchestration return a neutral result abstraction.
- Command-layer error unions can be simplified without dragging task-specific errors along.
- Future command-specific cleanups, including firmware metadata error cleanup, become much smaller and safer.

OlivierFreyssinet:
### 📝 Description

This PR extracts `DmkResult` as a generic base result abstraction for the DMK and rebuilds `CommandResult` as the command-specialized layer on top of it.

The goal is to stop non-command flows from using command-specific result helpers directly. In practice, this PR migrates DMK task and task-wrapper code (`GetFirmwareMetadataTask`, `GetApplicationsMetadataTask`, `BackupAppStorageTask`, `CallTaskInAppDeviceAction`, and `GetDeviceMetadataDeviceAction`) to `DmkResult` while keeping the command send/parse pipeline on `CommandResult`.

It also moves `InvalidGetFirmwareMetadataResponseError` out of the command layer and into task scope, so task-specific errors no longer pollute `CommandResult`.

A repo-wide lint policy is introduced as a warning on `@ledgerhq/device-management-kit` imports of `CommandResultFactory`, while DMK keeps a stricter local error for internal non-command production usage.

This PR mostly covers wrong usages inside the DMK package. The other packages still need to be tackled in order to fully close #1442.

This PR is stacked on top of #1441.
Follow-up tech debt issue: #1442.

### ❓ Context

- **JIRA or GitHub link**: `[NO-ISSUE]`, #1442
- **Feature**: Decouple generic workflow results from command-specific result typing

### ✅ Checklist

  - introduce `DmkResult`, `DmkResultFactory`, `DmkResultStatus`, and `isSuccessDmkResult` as the generic base abstraction
  - keep `CommandResult` and `CommandResultFactory` as the command-specialized compatibility layer
  - migrate DMK task flows and `CallTaskInAppDeviceAction` away from command-specific result construction
  - move `InvalidGetFirmwareMetadataResponseError` out of the command layer into task scope
  - add a shared warn-level lint policy and a stricter DMK-local lint guard against non-command `CommandResultFactory` usage
  - tighten type-level compatibility: code that constructs `CommandResultFactory` with `InvalidGetFirmwareMetadataResponseError` will now fail at the type level; non-command flows should use `DmkResultFactory` instead
  - update `signer-eth` and one `signer-zcash` test, covered by dedicated signer package changesets

### DIFF
--- a/.changeset/slow-owls-flow.md
+++ b/.changeset/slow-owls-flow.md
@@ -1,0 +1,7 @@
+---
+"@ledgerhq/device-management-kit": patch
+---
+
+Introduce `DmkResult` as a generic base result abstraction and move task-specific firmware metadata errors out of `CommandResult`.
+
+Code that constructs `CommandResultFactory` with `InvalidGetFirmwareMetadataResponseError` will now fail at the type level. Non-command flows should use `DmkResultFactory` instead.
--- a/.changeset/full-frogs-allow.md
+++ b/.changeset/full-frogs-allow.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/device-signer-kit-ethereum": patch
+---
+
+Align Ethereum device action error typing with the DMK task result error contract.
--- a/.changeset/salty-pumas-rush.md
+++ b/.changeset/salty-pumas-rush.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/device-signer-kit-zcash": patch
+---
+
+Align Zcash app binder task typing with the DMK task result error contract.
--- a/packages/config/eslint/index.js
+++ b/packages/config/eslint/index.js
@@ -134,6 +134,24 @@ export default [
       ],
     },
   },
+  {
+    files: ["**/*.ts", "**/*.tsx"],
+    ignores: ["**/*.test.ts", "**/*.test.tsx", "**/command/**"],
+    rules: {
+      "@typescript-eslint/no-restricted-imports": [
+        "warn",
+        {
+          paths: [
+            {
+              name: "@ledgerhq/device-management-kit",
+              importNames: ["CommandResultFactory"],
+              message: "Use DmkResultFactory outside the command layer.",
+            },
+          ],
+        },
+      ],
+    },
+  },
 
   {
     files: ["**/*.tsx"],
--- a/packages/device-management-kit/eslint.config.mjs
+++ b/packages/device-management-kit/eslint.config.mjs
@@ -10,4 +10,31 @@ export default [
       },
     },
   },
+  {
+    files: ["src/**/*.ts"],
+    ignores: [
+      "src/**/*.test.ts",
+      "src/api/command/**/*.ts",
+      "src/api/index.ts",
+    ],
+    rules: {
+      "@typescript-eslint/no-restricted-imports": [
+        "error",
+        {
+          paths: [
+            {
+              name: "@api/command/model/CommandResult",
+              importNames: ["CommandResultFactory"],
+              message: "Use DmkResultFactory outside the command layer.",
+            },
+            {
+              name: "@ledgerhq/device-management-kit",
+              importNames: ["CommandResultFactory"],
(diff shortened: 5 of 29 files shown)


.changeset/slow-owls-flow.md
```diff
@@ -0,0 +1,7 @@
+---
```
aussedatlo, line 1: [ASK] The PR description say that only DMK is modified, this only changeset too, but there is still edits on signer eth and zcash packages

OlivierFreyssinet: ah there's this part that explains why the changes don't need a changeset
> no additional changesets are needed there because these edits do not introduce new consumer-facing runtime behaviour or public API changes

OlivierFreyssinet: but yes the description isn't very clear

OlivierFreyssinet: In the end I added them so it's clear at the next release.

aussedatlo (approved):
Good job, the only issue I see there is that other devs can miss it, do you think it's worth adding documentation about this? or better, a skill?

PR #1450, opened 2026-04-27
✨ (ci) [NO-ISSUE]: Enforce single package per changeset in danger

aussedatlo:
### 📝 Description

Add a new Danger check that enforces the repo convention of **one package per changeset**. The check parses each `.changeset/*.md` file (excluding `README.md` and dotfiles) and fails the run if any of them declares more than one release entry.

This complements the existing guidance in `CONTRIBUTING.md` and `.cursor/rules/changeset.mdc`:

> Each changeset must affect only one package; create separate changesets for separate packages.

Also includes a couple of related improvements:

- Refactor `outputResults` to use Danger's `fail()` helper instead of `process.exit(1)`. The latter caused Danger to report a synthetic `node failed.` entry on top of our own report; using `fail()` marks the run as failed cleanly while keeping our generated markdown report as the single source of truth.
- Add `@changesets/parse` to the workspace catalog and root `devDependencies` so the check can rely on the official changeset parser.
- Split the existing `nasty-ideas-turn.md` changeset (which targeted both `@ledgerhq/context-module` and `@ledgerhq/device-signer-kit-ethereum`) into two single-package changesets so the new check passes on `develop` after this lands.

Example failure message (rendered in the Danger report):

> One or more changeset files declare more than one package. Each changeset must affect only one package, see CONTRIBUTING.md.
>
> **Offending changesets**:
> - `.changeset/foo-bar.md` declares 2 packages: `@ledgerhq/pkg-a`, `@ledgerhq/pkg-b`
>
> Please split the changeset above into one file per package.

### ❓ Context

- **Feature**: Tighten CI validation around changesets to prevent multi-package changesets from sneaking in and forcing manual splits at release time.

### ✅ Checklist

- [x] **Covered by automatic tests** — N/A, change is to the Danger CI tooling itself; the check is exercised on every PR run (and locally via `pnpm danger:local`).
- [ ] **Changeset is provided** — No changeset needed, changes only affect internal CI tooling (`danger/`) and root dev dependencies.
- [x] **Documentation is up-to-date** — `CONTRIBUTING.md` and `.cursor/rules/changeset.mdc` already document the single-package rule; this PR makes it enforceable.
  - New Danger check `checkChangesetSinglePackage` running on CI and locally for all PRs.
  - PRs adding a multi-package changeset will now fail Danger.
  - `outputResults` now requires a `fail` function argument (callers updated).
  - Adds `@changesets/parse` as a root dev dependency via the workspace catalog.
  - Splits `nasty-ideas-turn.md` into `nasty-ideas-turn.md` + `wise-mountains-read.md` (no behavior change for releases).

### DIFF
--- a/danger/helpers.ts
+++ b/danger/helpers.ts
@@ -2,7 +2,9 @@
 // so we need to pass them as arguments, only types can be imported
 import { type GitHubPRDSL, type DangerDSLType } from "danger";
 import { execSync } from "child_process";
-import { writeFileSync, appendFileSync } from "fs";
+import { writeFileSync, appendFileSync, readdirSync, readFileSync } from "fs";
+import { join } from "path";
+import parseChangeset from "@changesets/parse";
 
 // ── Types ──────────────────────────────────────────────────────────
 
@@ -53,7 +55,7 @@ const Branch = (danger: DangerDSLType, isFork: boolean = false) => ({
     ? new RegExp(`^(${BRANCH_PREFIX.join("|")})\/.+`, "i")
     : new RegExp(
         `^(release|chore\/backmerge(-.+){0,}|(${BRANCH_PREFIX.join("|")})\/((dsdk|live)-[0-9]+|no-issue|NOISSUE|issue-[0-9]+)-.+)`,
-        "i"
+        "i",
       ),
 
   getBranch: () => {
@@ -136,7 +138,7 @@ Special case for commit messages coming from a pull request merge:
 
     const currentBranch = Branch(danger, fork).getBranch();
     return execSync(
-      `git log origin/develop..${currentBranch} --pretty=format:%s`
+      `git log origin/develop..${currentBranch} --pretty=format:%s`,
     )
       .toString()
       .split("\n");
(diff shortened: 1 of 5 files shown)


danger/helpers.ts
```diff
@@ -401,8 +451,10 @@ export function outputResults(
...
-  // Exit with failure if any checks failed
-  if (checkResults.some((res) => !res.passed)) {
-    process.exit(1);
-  }
+  const failures = checkResults.filter((res) => !res.passed);
+  if (failures.length === 0) return;
+
+  // Use danger's `fail` so the run is marked failed without process.exit(1),
+  // which danger reports as a synthetic "node failed." entry in the PR comment.
+  fail(`Danger: ${failures.length} check(s) failed — see report above.`);
```
OlivierFreyssinet: Is this really necessary ? `markdown` already handles printing what's required, no?

aussedatlo: it's not but add a header showing that the job failed with x failures, it's a bit clearer when there is multiple errors

OlivierFreyssinet (approved):

PR #1521, opened 2026-05-27
✨ (device-management-kit) [NO-ISSUE]: Detect onboarded device from GetOsVersion

pdeville-ledger:
### 📝 Description

`GetDeviceStatusDeviceAction` was returning `isDeviceOnboarded: () => true`, so `DeviceNotOnboardedError` could never trigger and UIs (e.g. Ledger Button) had no way to detect an unseeded device.

This PR replaces the stub by reading `secureElementFlags.isOnboarded` from `GetOsVersionCommand` (no secure channel required), following the existing pattern in `ListInstalledAppsDeviceAction`.

- **`GetDeviceStatusDeviceAction`**:
  - `OnboardingCheck` reads cached `firmwareVersion.metadata` first; otherwise runs `GetOsVersionCommand` and branches on `isOnboarded`.
  - On success in a ready state, the session is updated with `firmwareVersion` (incl. `metadata`) and `isSecureConnectionAllowed`, so consumers reading `getDeviceSessionState()` get onboarding info without an extra command.
  - Removes the `isDeviceOnboarded` machine dependency / sync guard.
- **`OpenAppDeviceAction`**:
  - Removes its duplicate stubbed `OnboardingCheck`; relies on the fixed `GetDeviceStatus` child machine.
- **Tests**: mock `GetOsVersionCommand` instead of injecting `isDeviceOnboarded`; cover both the cached-metadata fast path and the live fetch path.

### ❓ Context

- **JIRA or GitHub link**: NO-ISSUE
- **Feature**: Real onboarded/seeded detection in DMK device actions

### ✅ Checklist

  - `GetDeviceStatusDeviceAction` may now return `Left(DeviceNotOnboardedError)` for unseeded devices.
  - `OpenAppDeviceAction` no longer emits its own pre-`GetDeviceStatus` onboarding step; consumers asserting on the state sequence may need updates.
  - `MachineDependencies` of both actions changed (`isDeviceOnboarded` removed; `getOsVersion` added on `GetDeviceStatus`).
  - Session state may now contain `firmwareVersion.metadata` and `isSecureConnectionAllowed` after `GetDeviceStatus` succeeds in a ready state.

### DIFF
--- a/packages/device-management-kit/src/api/device-action/os/GetDeviceStatus/GetDeviceStatusDeviceAction.ts
+++ b/packages/device-management-kit/src/api/device-action/os/GetDeviceStatus/GetDeviceStatusDeviceAction.ts
@@ -119,7 +119,7 @@ export class GetDeviceStatusDeviceAction extends XStateDeviceAction<
 
     const updateSessionFromOsVersion = (data: GetOsVersionResponse) => {
       const currentState = getDeviceSessionState();
-      if (currentState.sessionStateType === DeviceSessionStateType.Connected) {
+      if (!("firmwareVersion" in currentState)) {
         return;
       }
       setDeviceSessionState({
@@ -132,7 +132,7 @@ export class GetDeviceStatusDeviceAction extends XStateDeviceAction<
 
     const getCachedOnboardingMetadata = () => {
       const state = getDeviceSessionState();
-      if (state.sessionStateType === DeviceSessionStateType.Connected) {
+      if (!("firmwareVersion" in state)) {
         return undefined;
       }
       return state.firmwareVersion?.metadata;
--- a/packages/device-management-kit/src/api/device-action/os/GetDeviceStatus/GetDeviceStatusDeviceAction.ts
+++ b/packages/device-management-kit/src/api/device-action/os/GetDeviceStatus/GetDeviceStatusDeviceAction.ts
@@ -17,6 +17,11 @@ import {
   GetAppAndVersionCommand,
   type GetAppAndVersionCommandResult,
 } from "@api/command/os/GetAppAndVersionCommand";
+import {
+  GetOsVersionCommand,
+  type GetOsVersionCommandResult,
+  type GetOsVersionResponse,
+} from "@api/command/os/GetOsVersionCommand";
 import { DeviceStatus } from "@api/device/DeviceStatus";
 import { type InternalApi } from "@api/device-action/DeviceAction";
 import { UserInteractionRequired } from "@api/device-action/model/UserInteractionRequired";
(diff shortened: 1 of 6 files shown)


packages/device-management-kit/src/api/device-action/os/GetDeviceStatus/GetDeviceStatusDeviceAction.ts
```diff
@@ -93,14 +109,35 @@ export class GetDeviceStatusDeviceAction extends XStateDeviceAction<
...
+        isSecureConnectionAllowed:
+          data.secureElementFlags.isSecureConnectionAllowed,
+      });
+    };
+
+    const getCachedOnboardingMetadata = () => {
+      const state = getDeviceSessionState();
+      if (state.sessionStateType === DeviceSessionStateType.Connected) {
+        return undefined;
+      }
```
OlivierFreyssinet: I believe that two other `DeviceSessionType` have the metadata, i don't know if we actually use them, but in case we're in those states, you'll miss the data.
Maybe replace by `if "firmwareVersion" in state`.

packages/device-management-kit/src/api/device-action/os/GetDeviceStatus/GetDeviceStatusDeviceAction.ts
```diff
@@ -93,14 +109,35 @@ export class GetDeviceStatusDeviceAction extends XStateDeviceAction<
...
-      isDeviceOnboarded,
     } = this.extractDependencies(internalApi);
 
     const unlockTimeout = this.input.unlockTimeout ?? DEFAULT_UNLOCK_TIMEOUT_MS;
 
+    const updateSessionFromOsVersion = (data: GetOsVersionResponse) => {
+      const currentState = getDeviceSessionState();
+      if (currentState.sessionStateType === DeviceSessionStateType.Connected) {
+        return;
+      }
```
OlivierFreyssinet: this might be unnecessary and not optimal (for instance when in state `Connected`, you're not updating firmwareVersion even though it could be `undefined`.
IMO you can always override as your data here is the most up to date.

packages/device-management-kit/src/api/device-action/os/GetDeviceStatus/GetDeviceStatusDeviceAction.ts
```diff
@@ -131,10 +130,8 @@ export class GetDeviceStatusDeviceAction extends XStateDeviceAction<
     };
 
     const getCachedOnboardingMetadata = () => {
-      const state = getDeviceSessionState();
-      if (state.sessionStateType === DeviceSessionStateType.Connected) {
-        return undefined;
-      }
+      const state: DeviceReadyState =
+        getDeviceSessionState() as DeviceReadyState;
       return state.firmwareVersion?.metadata;
```
OlivierFreyssinet, line 138: unsafe type assertion, this opens the door to type errors

OlivierFreyssinet (approved):

PR #1530, opened 2026-06-02
✨ (signer-polkadot) [LIVE-30995]: Implement APDU commands for the Polkadot signer

cted-ledger:
### 📝 Description

Implement the full APDU command layer for the `signer-polkadot` DMK signer kit. The signer is generic across the Polkadot family (Polkadot, Westend, AssetHub, Bittensor, …) — chains differ only by the SS58 prefix parameter and the derivation path. Bittensor is the driving integration.

**Commands implemented:**
- **GetAddressCommand** — derives the address with a configurable SS58 prefix (0 = Polkadot, 42 = Bittensor) and optional on-device confirmation
- **SignTransactionCommand** — multi-chunk protocol (INIT/ADD/LAST) for signing transaction blobs, with the derivation path and blob length sent in the initial chunk
- **SignTransactionTask** — orchestrates blob+metadata chunking into sequential APDU calls, respecting the 255-byte payload limit

**Supporting utilities:**
- `EncodeDerivationPath` — encodes a hardened BIP44 path into the little-endian binary layout expected by the device app
- `polkadotApplicationErrors` — maps device status words to typed error codes
- Constants module with the app-wide CLA (INS/P2 live in their respective commands)

### ❓ Context

- **JIRA or GitHub link**: [LIVE-30995](https://ledgerhq.atlassian.net/browse/LIVE-30995)
- **Feature**: Polkadot-family signer kit (APDU layer) — driven by the Bittensor integration

Depends on #1501 (scaffold, merged).

### ✅ Checklist

- [x] **Covered by automatic tests** — unit tests across the command, task, use-case and signer layers (GetAddress/SignTransaction commands, SignTransactionTask, EncodeDerivationPath, application errors, GetAddressUseCase, DefaultSignerPolkadot, builder)
  - New APDU commands in `@ledgerhq/device-signer-kit-polkadot`
  - No impact on existing packages — all changes scoped to `signer-polkadot`
  - Multi-chunk signing protocol for large Substrate extrinsics

[LIVE-30995]: https://ledgerhq.atlassian.net/browse/LIVE-30995?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/signer/signer-polkadot/src/internal/app-binder/command/GetAddressCommand.ts
+++ b/packages/signer/signer-polkadot/src/internal/app-binder/command/GetAddressCommand.ts
@@ -1,11 +1,27 @@
 import {
   type Apdu,
+  ApduBuilder,
+  ApduParser,
   type ApduResponse,
   type Command,
   type CommandResult,
+  CommandResultFactory,
+  InvalidStatusWordError,
 } from "@ledgerhq/device-management-kit";
+import {
+  CommandErrorHelper,
+  DerivationPathUtils,
+} from "@ledgerhq/signer-utils";
+import { Maybe } from "purify-ts";
 
-import { type PolkadotErrorCodes } from "./utils/polkadotApplicationErrors";
+import {
+  POLKADOT_APP_ERRORS,
+  PolkadotAppCommandErrorFactory,
+  type PolkadotErrorCodes,
+} from "@internal/app-binder/command/utils/polkadotApplicationErrors";
+import { INS, LEDGER_CLA, P2 } from "@internal/app-binder/constants";
+
+import { encodeDerivationPath } from "./utils/EncodeDerivationPath";
 
 export type GetAddressCommandArgs = {
   readonly derivationPath: string;
@@ -15,8 +31,21 @@ export type GetAddressCommandArgs = {
 
 export type GetAddressCommandResponse = {
   readonly publicKey: Uint8Array;
-  readonly chainCode?: Uint8Array;
+  readonly address: string;
 };
+
+export const polkadotGetAddressApduHeader = (p1: number) => ({
+  cla: LEDGER_CLA,
+  ins: INS.GET_ADDRESS,
+  p1,
+  p2: P2.ED25519,
+});
+
+export const P1_CONFIRM = 0x01;
+export const P1_NO_CONFIRM = 0x00;
+
+const DERIVATION_PATH_LENGTH = 5;
+const PUBLIC_KEY_LENGTH = 32;
 
 export class GetAddressCommand
   implements
@@ -28,26 +57,71 @@ export class GetAddressCommand
 {
   readonly name = "GetAddress";
 
-  private readonly _args: GetAddressCommandArgs;
+  private readonly args: GetAddressCommandArgs;
+
+  private readonly errorHelper = new CommandErrorHelper<
+    GetAddressCommandResponse,
+    PolkadotErrorCodes
+  >(POLKADOT_APP_ERRORS, PolkadotAppCommandErrorFactory);
 
   constructor(args: GetAddressCommandArgs) {
-    this._args = args;
-  }
-
-  get args(): GetAddressCommandArgs {
-    return this._args;
+    this.args = args;
   }
 
   getApdu(): Apdu {
-    // TODO: Implement APDU construction using this._args
-    throw new Error("GetAddressCommand.getApdu() not implemented");
+    const apduBuilder = new ApduBuilder(
+      polkadotGetAddressApduHeader(
+        this.args.checkOnDevice ? P1_CONFIRM : P1_NO_CONFIRM,
+      ),
+    );
+
+    const derivationPath = DerivationPathUtils.splitPath(
+      this.args.derivationPath,
+    );
+
+    if (derivationPath.length !== DERIVATION_PATH_LENGTH) {
+      throw new Error(
+        `GetAddressCommand: expected ${DERIVATION_PATH_LENGTH} path elements, got ${derivationPath.length}`,
+      );
+    }
+
+    const encodedDerivationPath = encodeDerivationPath(derivationPath);
+    apduBuilder.addBufferToData(encodedDerivationPath);
+
+    // SS58 prefix: 2 bytes little-endian (firmware reads sizeof(uint16_t) via memcpy on ARM LE)
+    const ss58Buf = new Uint8Array(2);
+    new DataView(ss58Buf.buffer).setUint16(0, this.args.ss58Prefix, true);
+    apduBuilder.addBufferToData(ss58Buf);
+
+    return apduBuilder.build();
   }
 
   parseResponse(
-    _apduResponse: ApduResponse,
+    apduResponse: ApduResponse,
   ): CommandResult<GetAddressCommandResponse, PolkadotErrorCodes> {
-    // TODO: Implement response parsing based on your blockchain's protocol
-    // return CommandResultFactory({ data: { ... } });
-    throw new Error("GetAddressCommand.parseResponse() not implemented");
+    return Maybe.fromNullable(
+      this.errorHelper.getError(apduResponse),
+    ).orDefaultLazy(() => {
+      const apduParser = new ApduParser(apduResponse);
+      const publicKey = apduParser.extractFieldByLength(PUBLIC_KEY_LENGTH);
+      const remaining = apduParser.getUnparsedRemainingLength();
+      const address = apduParser.extractFieldByLength(remaining);
+
+      if (publicKey === undefined || address === undefined) {
+        return CommandResultFactory({
+          error: new InvalidStatusWordError("Public key is missing"),
+        });
+      }
+
+      // Address is variable-length ASCII, no null padding (unlike Cosmos)
+      const addressStr = apduParser.encodeToString(address);
+
+      return CommandResultFactory({
+        data: {
+          publicKey,
+          address: addressStr,
+        },
+      });
+    });
   }
 }
--- a/packages/signer/signer-polkadot/src/internal/app-binder/command/GetAddressCommand.ts
+++ b/packages/signer/signer-polkadot/src/internal/app-binder/command/GetAddressCommand.ts
@@ -19,9 +19,14 @@ import {
   PolkadotAppCommandErrorFactory,
   type PolkadotErrorCodes,
 } from "@internal/app-binder/command/utils/polkadotApplicationErrors";
-import { INS, LEDGER_CLA, P2 } from "@internal/app-binder/constants";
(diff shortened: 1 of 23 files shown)


packages/signer/signer-polkadot/src/internal/app-binder/command/utils/polkadotApplicationErrors.ts
```diff
@@ -1 +1,45 @@
...
+  CLA_NOT_SUPPORTED = "0x6E00",
+  UNKNOWN_ERROR = "0x6F00",
+  SIGN_VERIFY_ERROR = "0x6F01",
+}
+
+export const POLKADOT_APP_ERRORS: CommandErrors<PolkadotErrorCodes> = {
+  "0x6400": { message: "Execution Error" },
+  "0x6700": { message: "Wrong buffer length" },
+  "0x6982": { message: "Empty buffer" },
+  "0x6983": { message: "Output buffer too small" },
```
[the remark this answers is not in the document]
cted-ledger: fixed

packages/signer/signer-polkadot/src/internal/app-binder/constants.ts
```diff
@@ -7,3 +7,8 @@ export const INS = {
   SIGN_TRANSACTION: 0x02,
   SIGN_RAW: 0x03,
 } as const;
+
+export const P2 = {
```
mbertin-ledger: [SHOULD] P2 is parameter for a given command, this should be directly in command `SignTransactionCommand` as it's used there. 

Same for INS above.

mbertin-ledger (changes requested):
Please Handle also comment from sonarcloud analysis

packages/signer/signer-polkadot/src/internal/app-binder/constants.ts
cted-ledger: Done in 251691e6 — moved the command-specific INS and P2 into GetAddressCommand and SignTransactionCommand; constants.ts now keeps only the app-wide LEDGER_CLA (also dropped the unused SIGN_RAW/ECDSA).

packages/signer/signer-polkadot/src/internal/app-binder/command/GetAddressCommand.ts
```diff
@@ -28,26 +57,71 @@ export class GetAddressCommand
...
+      );
+    }
+
+    const encodedDerivationPath = encodeDerivationPath(derivationPath);
+    apduBuilder.addBufferToData(encodedDerivationPath);
+
+    // SS58 prefix: 2 bytes little-endian (firmware reads sizeof(uint16_t) via memcpy on ARM LE)
+    const ss58Buf = new Uint8Array(2);
+    new DataView(ss58Buf.buffer).setUint16(0, this.args.ss58Prefix, true);
+    apduBuilder.addBufferToData(ss58Buf);
```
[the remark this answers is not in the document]
cted-ledger: Fixed in 251691e6 — getApdu now throws if ss58Prefix is not an integer in [0, 0xFFFF] before the uint16 LE write (test added).

packages/signer/signer-polkadot/src/internal/app-binder/command/SignTransactionCommand.ts
```diff
@@ -25,26 +60,93 @@ export class SignTransactionCommand
...
+      );
+    }
+
+    const encodedDerivationPath = encodeDerivationPath(paths);
+    this.apduBuilder.addBufferToData(encodedDerivationPath);
+
+    // blobLength: 2 bytes little-endian (firmware reads sizeof(uint16_t) via memcpy on ARM LE)
+    const blobLenBuf = new Uint8Array(2);
+    new DataView(blobLenBuf.buffer).setUint16(0, blobLength, true);
+    this.apduBuilder.addBufferToData(blobLenBuf);
```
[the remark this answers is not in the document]
cted-ledger: Fixed in 251691e6 — firstChunk now throws if blobLength exceeds the uint16 range before the LE write (test added).

packages/signer/signer-polkadot/src/internal/app-binder/task/SignTransactionTask.ts
```diff
@@ -19,29 +26,84 @@ export class SignTransactionTask {
...
+      return CommandResultFactory({
+        error: new InvalidStatusWordError("Failed to sign transaction"),
+      });
+    }
+
+    // Loop through payload in APDU_MAX_PAYLOAD chunks
+    for (let offset = 0; offset < payload.length; offset += APDU_MAX_PAYLOAD) {
+      const isLastChunk = offset + APDU_MAX_PAYLOAD >= payload.length;
+      const phase = isLastChunk ? SignPhase.LAST : SignPhase.ADD;
+      const transactionChunk = payload.slice(offset, offset + APDU_MAX_PAYLOAD);
```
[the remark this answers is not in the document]
cted-ledger: The PR description now states 255 bytes, matching APDU_MAX_PAYLOAD used for chunking — consistent.

packages/signer/signer-polkadot/src/internal/app-binder/command/SignTransactionCommand.ts
```diff
@@ -25,26 +66,103 @@ export class SignTransactionCommand
...
+
+  private readonly errorHelper = new CommandErrorHelper<
+    SignTransactionCommandResponse,
+    PolkadotErrorCodes
+  >(POLKADOT_APP_ERRORS, PolkadotAppCommandErrorFactory);
 
   constructor(args: SignTransactionCommandArgs) {
-    this._args = args;
+    this.args = args;
+    this.apduBuilder = new ApduBuilder(
```
fAnselmi-Ledger: [MUST] `apduBuilder` should be a local variable created inside `getApdu()` on every call not an instance field, you can take a look at `GetAddressCommand`

packages/signer/signer-polkadot/src/internal/app-binder/task/SignTransactionTask.ts
```diff
@@ -19,29 +26,84 @@ export class SignTransactionTask {
...
       }),
     );
 
-    if (!isSuccessCommandResult(result)) {
-      return result;
+    if (!isSuccessCommandResult(initResult)) {
+      this.logger.debug("[run] Failed to sign transaction", {
+        data: { error: initResult.error },
+      });
+      return CommandResultFactory({
```
fAnselmi-Ledger: [ASK] shouldn't this be consistent with the error propagation implemented below (ln99)?

packages/signer/signer-polkadot/src/internal/app-binder/command/SignTransactionCommand.ts
```diff
@@ -25,26 +66,103 @@ export class SignTransactionCommand
...
+  ): CommandResult<SignTransactionCommandResponse, PolkadotErrorCodes> {
+    return Maybe.fromNullable(
+      this.errorHelper.getError(apduResponse),
+    ).orDefaultLazy(() => {
+      const apduParser = new ApduParser(apduResponse);
+      const remaining = apduParser.getUnparsedRemainingLength();
+      const signature = apduParser.extractFieldByLength(remaining);
+
+      return CommandResultFactory({
+        data: signature as Signature,
```
fAnselmi-Ledger: [ASK] you shouldn't be forced to cast `as Signature`?

packages/signer/signer-polkadot/src/internal/app-binder/task/SignTransactionTask.ts
```diff
@@ -19,29 +26,84 @@ export class SignTransactionTask {
...
+      }
+
+      if (isLastChunk) {
+        this.logger.debug("[run] Transaction signed successfully", {
+          data: {
+            signature: result.data,
+          },
+        });
+        return CommandResultFactory({
+          data: result.data as Signature,
```
fAnselmi-Ledger: [ASK] you shouldn't be forced to cast `as Signature`?

packages/signer/signer-polkadot/src/internal/app-binder/command/SignTransactionCommand.ts
cted-ledger: Fixed in 1ea13871 — the `ApduBuilder` is now created inside `getApdu()` and handed to `firstChunk()`/`nextChunk()`, matching `GetAddressCommand`.

Worth noting it was not only a style issue: as an instance field the builder accumulated into its data, so a second `getApdu()` call on the same instance produced an APDU with the payload appended twice. Two tests now assert that calling `getApdu()` twice yields the same APDU (INIT and ADD).

packages/signer/signer-polkadot/src/internal/app-binder/task/SignTransactionTask.ts
cted-ledger: You're right, it was inconsistent. The INIT failure replaced the device error with a generic `InvalidStatusWordError("Failed to sign transaction")`, while the chunk loop below returns `result` untouched — so a user rejection or a `DATA_INVALID` on INIT was indistinguishable from any other failure.

Fixed in 1ea13871: it now returns `initResult` directly. The test asserts the `PolkadotAppCommandError` propagates unchanged, the same way a failing data chunk does.

packages/signer/signer-polkadot/src/internal/app-binder/command/SignTransactionCommand.ts
cted-ledger: Correct, and the cast was hiding a real case. `extractFieldByLength` returns `Uint8Array | undefined`, and `extractFieldByLength(0)` returns an *empty array* rather than `undefined` (`ByteArrayParser.ts:149`) — so a `9000` with an empty body parsed as a valid, empty signature.

Fixed in 1ea13871: the cast is gone and the LAST phase now requires exactly 65 bytes, which is what the device always replies (`SIG_PLUS_TYPE_LEN` = 1-byte MultiSignature discriminant + 64-byte Ed25519 signature). The check is scoped to LAST because INIT and ADD legitimately answer with an empty body — the app returns `APDU_CODE_OK` with no data until the last chunk.

packages/signer/signer-polkadot/src/internal/app-binder/task/SignTransactionTask.ts
cted-ledger: Agreed, this one was simply unnecessary: `result.data` is already typed `SignTransactionCommandResponse`, which is `Signature`. Replaced by `return result;` in 1ea13871, and the now-unused `Signature` import is gone.

fAnselmi-Ledger (approved):

PR #1532, opened 2026-06-03
✨ (device-management-kit) [NO-ISSUE]: Add WaitAppAndVersionDeviceAction

benruseau:
### 📝 Description

This PR introduces a new **`WaitForAppAndVersionDeviceAction`** in the Device Management Kit.

> **Note:** `GetDeviceStatusDeviceAction` currently reimplements this same unlock-polling logic internally. A follow-up could refactor it to reuse `WaitForAppAndVersionDeviceAction`, removing the duplicated polling code and keeping a single source of truth for the "get app & version / wait for unlock" behavior.

**The why:**

We need a device action that retrieves the currently running app **without requiring the device to be onboarded**, and that can be used as a building block to reach the dashboard for OS-level operations (e.g. the upcoming **OS update migration**, which needs to run a `GetOsVersion` command on the dashboard).

Today our only options have important constraints:

- **`GetDeviceStatusDeviceAction`** rejects non-onboarded devices: it runs an onboarding check and fails with `DeviceNotOnboardedError`. We don't want that here — **OS updates must also work on devices that are not yet onboarded.**
- **`GoToDashboardDeviceAction`** can't be used as the first step for the same reason: it relies on device-status / onboarding assumptions and will fail on a non-onboarded device.

`GetAppAndVersion` is the right primitive because it can be issued safely on any device state (onboarded or not, dashboard or app), only requiring the device to be unlocked. It lets us answer "are we on the dashboard?", then if not (if the device is not onboarded, it will), run `GoToDashboardDeviceAction` and drive the OS-update path accordingly.

image

**Polling mechanism:**

The unlock-wait/polling logic is **inspired by `GetDeviceStatusDeviceAction`** but **extracted into a dedicated, single-purpose device action**. Rather than embedding the full device-status / onboarding flow, we reuse only the lightweight `interval` + `GetAppAndVersion` polling pattern (with a configurable `unlockTimeout`, defaulting to `DEFAULT_UNLOCK_TIMEOUT_MS`). This keeps the action reusable and free of the onboarding constraint, so it can serve as the entry point for the OS update path.
### ❓ Context

- **JIRA or GitHub link**: NO TICKET

### ✅ Checklist

- [x] **Covered by automatic tests** — `WaitForAppAndVersionDeviceAction.test.ts` covers success, locked → unlock polling, timeout, and error cases.
  - Adds a new public device action exported from `@api`.
  - QA focus: locked-device unlock polling + timeout behavior, and behavior on **non-onboarded** devices.

### DIFF
--- a/.changeset/grumpy-pants-switch.md
+++ b/.changeset/grumpy-pants-switch.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/device-management-kit": minor
+---
+
+Add WaitForAppAndVersionDeviceAction
--- a/packages/device-management-kit/src/api/device-action/os/WaitForAppAndVersion/WaitForAppAndVersionDeviceAction.test.ts
+++ b/packages/device-management-kit/src/api/device-action/os/WaitForAppAndVersion/WaitForAppAndVersionDeviceAction.test.ts
@@ -1,0 +1,292 @@
+import { of, throwError } from "rxjs";
+
+import { CommandResultFactory } from "@api/command/model/CommandResult";
+import {
+  GLOBAL_ERRORS,
+  GlobalCommandError,
+} from "@api/command/utils/GlobalCommandError";
+import { makeDeviceActionInternalApiMock } from "@api/device-action/__test-utils__/makeInternalApi";
+import { testDeviceActionStates } from "@api/device-action/__test-utils__/testDeviceActionStates";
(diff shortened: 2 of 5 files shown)


smartine-ledger (approved):
LGTM

OlivierFreyssinet (approved):

PR #1583, opened 2026-06-29
✨ (devtools) [NO-ISSUE]: Add devtools session to mocks

benruseau:
### 📝 Description

This PR introduces a **Mock Capture** screen in the DevTools dashboard that turns a real device session into a ready-to-replay mock-server configuration.

While a device performs operations (connect, open app, sign, etc.), the DevTools logger passively records every APDU exchange (request + response) coming from the host app's DMK logs. The screen then derives mock-server mocks from those exchanges and the live session metadata, and lets you **download the session export as JSON** or **push it directly to a running mock server** for replay.

To make this reliable, the DMK now also emits a **structured APDU exchange log** (request, response, and session id) so consumers can reconstruct exchanges from the log payload instead of fragile string parsing.

**Key behaviors / advantages:**

- **Auto-derived device metadata**: commercial model name (`Flex`, `Stax`, `Gen 5`, …), `device_type`, `connectivity_type` (`USB`/`BLE`), and SE `firmware_version` are inferred from the live session.
- **Resilient firmware detection**: firmware version is read while on the dashboard (BOLOS) and persisted, so it isn't erased when you later open an app.
- **App accumulation**: switching apps automatically adds to the `apps` list; the dashboard (`BOLOS`) is filtered out as it is not an installed app.
- **Multi-device support**: when several devices are connected, exchanges are grouped per session so each device gets its own entry and mocks (no cross-device collision).
- **Optional handshake mocks**: a toggle includes/excludes the `GetOsVersion`/`GetAppAndVersion` handshake APDUs.
- **Editable JSON**: the export is shown in an editable black box; you can tweak it directly before download/push (with live JSON validation).
- **Direct push to mock server**: push a generated session directly to a running mock server.

**Result:**

https://github.com/user-attachments/assets/cd60043b-323a-41ca-89f1-9f24c8b51660

### ❓ Context

- **Feature**: Capture a live device session and convert it into mock-server mocks directly from the DevTools dashboard, streamlining the creation of deterministic, replayable test fixtures.

### ✅ Checklist

- [x] **Covered by automatic tests** — unit tests for the structured APDU exchange log (`apduLogs.test.ts`) and for session-export generation, including multi-device grouping and "name first" ordering (`generateSessionExport.test.ts`). The React screen (`index.tsx`) itself is not yet covered by component tests.
- [x] **Changeset is provided** — `@ledgerhq/device-management-kit-devtools-ui` (minor) and `@ledgerhq/device-management-kit` (patch).
  - New `Mock Capture` screen + nav entry in `@ledgerhq/device-management-kit-devtools-ui`.
  - New structured `apdu-exchange` log emitted by `DeviceSession` in `@ledgerhq/device-management-kit` (additive; existing string logs unchanged).
  - DevTools Electron CSP updated to permit local mock-server HTTP requests.
  - QA focus: recording start/stop/clear, JSON edit + validation, download, push to mock server (auth + session token paths), USB vs BLE connectivity, firmware detection across app switches, and multi-device capture.

### DIFF
--- a/packages/device-management-kit/src/api/utils/apduLogs.ts
+++ b/packages/device-management-kit/src/api/utils/apduLogs.ts
@@ -25,3 +25,29 @@ export function formatApduSentLog(apdu: Uint8Array): string {
 export function formatApduReceivedLog(apduResponse: ApduResponse): string {
   return `[exchange] <= ${bufferToHexaString(apduResponse.data, false)}${bufferToHexaString(apduResponse.statusCode, false)}`;
 }
+
+export const APDU_EXCHANGE_LOG = "apdu-exchange";
+
+export type ApduExchangeLog = {
+  type: typeof APDU_EXCHANGE_LOG;
+  sessionId: string;
+  apdu: string;
+  response: string;
+};
+
+/**
+ * Formats the log message and builds the structured `data` payload for a completed APDU exchange
+ * (request + response).
+ */
+export function formatApduExchangeLog(
+  sessionId: string,
+  apdu: Uint8Array,
+  apduResponse: ApduResponse,
+): { message: string; data: ApduExchangeLog } {
+  const request = bufferToHexaString(apdu, false);
+  const response = `${bufferToHexaString(apduResponse.data, false)}${bufferToHexaString(apduResponse.statusCode, false)}`;
+  return {
+    message: formatApduReceivedLog(apduResponse),
+    data: { type: APDU_EXCHANGE_LOG, sessionId, apdu: request, response },
+  };
+}
--- a/packages/device-management-kit/src/internal/device-session/model/DeviceSession.ts
+++ b/packages/device-management-kit/src/internal/device-session/model/DeviceSession.ts
@@ -31,7 +31,7 @@ import {
 } from "@api/transport/model/Errors";
 import { type TransportConnectedDevice } from "@api/transport/model/TransportConnectedDevice";
 import {
-  formatApduReceivedLog,
+  formatApduExchangeLog,
   formatApduSendingLog,
 } from "@api/utils/apduLogs";
 import { DEVICE_SESSION_REFRESHER_DEFAULT_OPTIONS } from "@internal/device-session/data/DeviceSessionRefresherConst";
@@ -246,7 +246,12 @@ export class DeviceSession {
 
     return result
       .ifRight((response: ApduResponse) => {
-        this._logger.debug(formatApduReceivedLog(response));
+        const { message, data } = formatApduExchangeLog(
+          this._id,
+          rawApdu,
+          response,
+        );
+        this._logger.debug(message, { data });
         if (CommandUtils.isLockedDeviceResponse(response)) {
           this._sessionEventDispatcher.dispatch({
             eventName: SessionEvents.DEVICE_STATE_UPDATE_LOCKED,
--- a/packages/devtools/ui/src/hooks/connectorMessageHandlers.ts
+++ b/packages/devtools/ui/src/hooks/connectorMessageHandlers.ts
@@ -8,7 +8,34 @@ import { type DevToolsModule } from "@ledgerhq/device-management-kit-devtools-co
 
 import { mapConnectorMessageToLogData } from "../screens/logger/mapConnectorMessageToLogData";
 import { type LogData } from "../screens/logger/types";
-import { type ApduResponse } from "./useConnectorMessages";
+import { type ApduExchange, type ApduResponse } from "./useConnectorMessages";
+
+const APDU_EXCHANGE_LOG = "apdu-exchange";
+
+/**
+ * Extracts a structured APDU exchange from a log message, if the log carries
+ * the `apdu-exchange` payload emitted by DMK. Returns null otherwise.
+ */
+function tryExtractApduExchange(logData: LogData): ApduExchange | null {
+  const data = logData.payload;
+  if (
+    typeof data !== "object" ||
+    data === null ||
+    Array.isArray(data) ||
+    data["type"] !== APDU_EXCHANGE_LOG ||
+    typeof data["apdu"] !== "string" ||
+    typeof data["response"] !== "string"
+  ) {
+    return null;
+  }
+  return {
+    sessionId:
+      typeof data["sessionId"] === "string" ? data["sessionId"] : undefined,
+    apdu: data["apdu"],
+    response: data["response"],
+    timestamp: logData.timestamp,
+  };
+}
 
 /**
  * Handle module connected handshake message.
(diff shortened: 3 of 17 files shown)


packages/device-management-kit/src/internal/device-session/model/DeviceSession.ts
```diff
@@ -246,7 +246,12 @@ export class DeviceSession {
 
     return result
       .ifRight((response: ApduResponse) => {
-        this._logger.debug(formatApduReceivedLog(response));
+        const { message, data } = formatApduExchangeLog(
+          this._id,
+          rawApdu,
+          response,
+        );
+        this._logger.debug(message, { data });
```
OlivierFreyssinet, line 254: ⚠️ breaking change

We cannot remove the log "[exchange] <= 9000" logs as external tools rely on it to extract logs, and those tools are used by the embedded team to troubleshoot issues.

So I see 3 possibilities:
- adapt the logs extraction logic to work without the new formatting (not recommended as it's easy to break, versions will need to be aligned)
- adapt your work so it works with the existing apdu logging logic (I think it's doable)
- keep both "apdu received" and the new "apdu exchange" log (a bit redundant, but why not)

OlivierFreyssinet (changes requested):

packages/device-management-kit/src/internal/device-session/model/DeviceSession.ts
benruseau: My bad! 😨 

I refactored `formatApduExchangeLog` to keep the exact same log message with the attached data so it can still be extracted by the devtools at the same time:  https://github.com/LedgerHQ/device-sdk-ts/pull/1583/changes#diff-95896dce50d56b5e481cca3eb666655db11ae4373012828e80530b1f83759ebeR42

packages/devtools/ui/src/screens/mock-capture/deviceConfigFromSession.ts
```diff
@@ -0,0 +1,77 @@
...
+  if (!state) {
+    return {};
+  }
+
+  const firmwareVersion = firmwareVersionFromState(state);
+
+  const base: Partial<DeviceConfig> = {
+    ...(DEVICE_MODEL_NAMES[state.deviceModelId]
+      ? { name: DEVICE_MODEL_NAMES[state.deviceModelId] }
+      : {}),
```
OlivierFreyssinet: [COULD] I think here it's better to do something with `StaticDeviceModelDataSource` to obtain the product name rather than reimplementing a record

packages/devtools/ui/src/hooks/useConnectorMessages.ts
```diff
@@ -162,6 +182,13 @@ export function useConnectorMessages(
...
           // Try to handle as log message
           handleLogMessage(type, payload, setLogs);
+          // Passively capture APDU exchanges from logs while recording
+          if (isRecordingExchangesRef.current) {
+            const exchange = tryExtractApduExchange(type, payload);
+            if (exchange !== null) {
+              setApduExchanges((prev) => [...prev, exchange]);
+            }
+          }
           break;
```
OlivierFreyssinet: [SHOULD] Can this be done in `handleLogMessage` for consistency and to keep this file flat? (you'll have to pass it `setApduExchanges` and `isRecordingExchangesRef.current`).

By the way I think it's an opportunity to clean the `switch` block, use a `case LOGGER_MESSAGE_TYPES.ADD_LOG:` instead of the `default` path, so type does not have to be passed to `handleLogMessage`. Would appreciate that 🙏

OlivierFreyssinet (changes requested):
Thanks for the fix 🙏
Can you just do a tiny refacto in `useConnectorMessages` please to keep it clean

benruseau:
> Thanks for the fix 🙏 Can you just do a tiny refacto in `useConnectorMessages` please to keep it clean

Should be ok :)

packages/device-management-kit/src/api/utils/apduLogs.ts
```diff
@@ -25,3 +25,29 @@ export function formatApduSentLog(apdu: Uint8Array): string {
...
+  sessionId: string;
+  apdu: string;
+  response: string;
+};
+
+/**
+ * Formats the log message and builds the structured `data` payload for a completed APDU exchange
+ * (request + response).
+ */
+export function formatApduExchangeLog(
```
smartine-ledger, line 42: [ASK] Do we really want this in the dmk? I mean isn't it just for the devTool?

benruseau: Good question. Indeed it is consumed by the devtools, but I think the emission belongs in DMK because DMK is the only layer that has the completed exchange with sessionId, request APDU and corresponding response data at the same point in time. The devtools should only consume/record that structured log.

Doing everything on the devtools side would be much more complex I guess because we would need to correlate independent log entries, keep per-session pending APDUs, handle interleaved polling/user APDUs and stay coupled to the exact text log format. Emitting one structured exchange from DMK also avoids that fragile reconstruction.

That said, it’s not meant to become public in the DMK API.

smartine-ledger: Agree but it looks like you're formatting data specifically for the devtool inside the DMK IIUC. IMHO the DMK should just emit generic logs and each consumer should format them as needed. But I don't know if it's in the scope of your PR 🤷‍♂️

benruseau: Yes, and that’s pretty much what this does.

The emitted log message stays the same as before: it’s still the generic APDU response log. The only change is that we attach structured metadata to that log entry through its optional `options` field so the consumers can reconstruct the APDU exchange context when needed: sessionId, associated request APDU, response, etc.

Even if the devtools is currently the only consumer using this metadata, I think it still fits DMK’s responsibility because DMK is the only place that reliably has this context at emission time. Consumers remain free to ignore the metadata or format it however they want.

wdyt?

smartine-ledger: Ok then 😉 . LGTM

smartine-ledger (approved):

packages/devtools/ui/src/hooks/connectorMessageHandlers.ts
```diff
@@ -29,13 +69,23 @@ export function handleModuleConnected(
...
 export function handleLogMessage(
-  type: string,
   payload: string,
   setLogs: Dispatch<SetStateAction<LogData[]>>,
+  setApduExchanges: Dispatch<SetStateAction<ApduExchange[]>>,
+  isRecordingExchanges: boolean,
 ): boolean {
-  const logData = mapConnectorMessageToLogData({ type, payload });
+  const logData = mapConnectorMessageToLogData({
+    type: LOGGER_MESSAGE_TYPES.ADD_LOG,
```
OlivierFreyssinet: this type param isn't required anymore (this is the only call site), it was just there because in the `switch` statement we were using the `default` branch.

packages/devtools/ui/src/hooks/connectorMessageHandlers.ts
```diff
@@ -4,11 +4,51 @@ import {
...
+    if (!log.payloadJSON) {
+      return null;
+    }
+    const data = JSON.parse(log.payloadJSON) as Partial<{
+      type: string;
+      sessionId: string;
+      apdu: string;
+      response: string;
+    }>;
+    if (
```
OlivierFreyssinet: Normally you don't have to parse this data again as it's already done by `mapConnectorMessageToLogData` and the data you need is all in `LogData`.
So this function should just take `logData: LogData` as a param.

OlivierFreyssinet (approved):

PR #1592, opened 2026-07-02
🎨 (signer-solana) [DSDK-1310]: Align with new OCMS v1 specs (drop length)

fAnselmi-Ledger:
### 📝 Description

This aligns Solana off-chain message signing (OCMS) V1 with the finalised sRFC 38 spec, which drops the 2-byte little-endian message length prefix. In the new layout the message body is simply the trailing bytes after the signers, with no length field.

To stay compatible, the sign-message task now cascades through the possible layouts and advances on the device 6a81 (invalid off-chain header) reject, which is the same mechanism the existing V0 to Legacy fallback already uses. The order is:
  1. V1 without length prefix (finalised sRFC 38, preferred)
  2. V1 with length prefix (pre-spec-update firmware)
  3. V0
  4. Legacy

**Additional changes in this PR:**
  - Message length cap: V0 and V1 now use the real device limit that app-solana enforces (`MAX_OFFCHAIN_MESSAGE_LENGTH`, which is `PACKET_DATA_SIZE`, that is (15 times 1024) minus 40 minus 8, so 15312 bytes). Previously the caps were 65515 and 65535
  - Fallback logging: Each cascade transition now logs through the standard logger (warn on each fallback, error when every layout is exhausted), so the accepted layout can be traced from the exported logs.

### ❓ Context

- **JIRA or GitHub link**: [DSDK-1310](https://ledgerhq.atlassian.net/browse/DSDK-1310)

### ✅ Checklist

  - _list of the changes_

---

### DIFF
--- a/packages/signer/signer-solana/src/internal/app-binder/task/SendSignMessageTask.ts
+++ b/packages/signer/signer-solana/src/internal/app-binder/task/SendSignMessageTask.ts
@@ -5,6 +5,7 @@ import {
   InvalidArgumentError,
   InvalidStatusWordError,
   isSuccessCommandResult,
+  type LoggerPublisherService,
 } from "@ledgerhq/device-management-kit";
 import {
   type CommandFactory,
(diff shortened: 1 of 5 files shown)


smartine-ledger (commented):
✅. However IMHO the `SendSignMessageTask` does too much and should probably be split into several different tasks

packages/signer/signer-solana/src/internal/app-binder/task/SendSignMessageTask.ts
```diff
@@ -222,15 +227,31 @@ export class SendSignMessageTask {
...
+
+    // Preferred: finalised sRFC 38 layout (no length prefix). Older firmware
+    // rejects it with 6a81 (header.length != trailing bytes), so fall back to
+    // the pre-spec-update layout that still carries the 2-byte length prefix.
+    const v1OCM = this._builder.buildV1(sendingData, signers, false);
     const v1Result = await this._sendAndWrap(v1OCM, paths);
+    if (v1Result) {
+      return v1Result;
+    }
+    this._logger?.warn(
```
aussedatlo, line 240: in this place how can you be sure that the device returned a `6a81` error without looking at the result? could it be another error code?

aussedatlo (approved):

PR #1623, opened 2026-07-08
🍪 feat (cs-tester) [NO-ISSUE]: Use speculinho on Clear-Signing Tester

pvautherin-ledger:
## Summary

Migrates the Ethereum and Solana nightly clear-signing tests from a local Docker/Speculos setup to **Speculinho** (remote device pods), and adds several robustness improvements to HTTP data sources and test fixtures.

### Key changes

- **Docker → Speculinho**: Remove `NodeDockerContainer`, `SpeculosServiceController`, `AppVersionResolverService`, and all Docker orchestration code. Replace with `SpeculinhoServiceController` that connects to a remote Speculinho URL. App and OS versions are now explicitly provided via CI inputs rather than resolved at runtime from a Docker image.
- **HTTP retry logic**: Add `withRetry` utility and apply it to `HttpGatedDescriptorDataSource` and `HttpSafeProxyDataSource` to make flaky network calls resilient in CI.
- **Web3 Checks / opt-out screen detection**: Add `OptOutStateHandler` and `ScreenAnalyzer` to handle the opt-in/opt-out screens that some app versions display before the transaction review.
- **Navigation timing**: Improve `SignTransactionStateHandler` and `HttpSpeculosDatasource` to handle slower real-device response times on Speculinho vs. local Docker.
- **Gating test fixtures**: Restore `expectedTexts` on transaction 1/11 in `raw-gating-every-10-tx.json` (was accidentally dropped in a simplification commit) and align `raw-multisig.json` with the correct gating screen wording.

---

### Tests run

- ETH: https://github.com/LedgerHQ/device-sdk-ts/actions/runs/29408805230 ✅ 
- SOL: https://github.com/LedgerHQ/device-sdk-ts/actions/runs/29408805284 ✅ 

---

### Undocumented caveats discovered during development

These were not obvious from the code and are documented here for future reference.

**1. The old Docker Speculos image bundled a pre-1.22.0 Ethereum app**

The previous Docker-based setup pulled a Speculos image that included an old Ethereum app (pre-1.22.0). That app version did not support calldata clear-signing for Safe multisig transactions, so it always fell through to the gating screen. This made the gating tests pass "for the wrong reason" — the test was verifying old app behavior, not the access-control mechanism.

**2. Ethereum app >= 1.22.0 changes gating screen behavior**

Starting with Eth app 1.22.0, when a calldata descriptor is available, the device shows a full clear-signed review (`"Interaction with Safe{Wallet}"`, `"Ledger Multisig"`) instead of the gating screen. Whether a descriptor is returned depends on the `Ledger-Origin-Token` header sent to CAL (see caveat 3).

**3. `--skip-origin-token` is intentional in gating tests**

`test-gating.sh` always passes `--skip-origin-token`, which omits the `Ledger-Origin-Token` header from CAL requests. CAL withholds calldata descriptors for unauthenticated clients — even on Eth app >= 1.22.0 — causing the device to fall back to the gating screen. This is the correct way to test the gating mechanism regardless of app version. Running the same test *without* this flag (e.g. manually, with a `GATING_TOKEN` in the environment) will show a full calldata review instead, which is confusing but expected.

**4. "Gating" ≠ "authenticating" — it's a commercial redirect**

The gating screen (`"There is a safer way to sign"`) is not a security gate. It is a product UX that redirects users toward Ledger Enterprise Multisig when they attempt to sign a Safe transaction without a calldata descriptor. The test verifies this redirect happens for unauthenticated callers.

**5. `HttpSafeProxyDataSource` returning 404/500 is expected for factory contracts**

During test runs, `HttpSafeProxyDataSource` logs errors for certain contract addresses. These are not regressions — the Safe ProxyFactory contract (`0x4e1d...`) is not itself a proxy, so a proxy lookup on it correctly returns nothing. The retry logic now makes these failures more visible but they were always silently occurring.

**6. `raw-gating-every-10-tx.json`: both tx 1 and tx 11 must have `expectedTexts`**

The Ethereum app firmware only shows the gating screen once every ~10 transactions (to avoid annoying the user). The file models this: tx 1 triggers the gating screen, tx 2–10 pass through silently, tx 11 triggers it again.


packages/device-management-kit/src/api/network/DmkNetworkClient.ts
```diff
@@ -206,7 +206,23 @@ export class DmkNetworkClient {
...
+        // `fetch()` can resolve with headers before the body finishes
+        // streaming (e.g. chunked responses, or servers that hold the
+        // connection open while the caller is blocked on some external
+        // event). If the underlying socket is then closed/reset while the
+        // body is still being read, that failure surfaces here — not from
+        // the initial `fetch()` call above — so it must be wrapped the same
+        // way, otherwise callers can't distinguish it from a JSON parsing
+        // error and it won't be recognized as a connectivity failure.
+        if (cause instanceof DmkNetworkClientError) {
+          throw cause;
```
pvautherin-ledger: fetch can resolve with headers before the body finishes
        streaming (e.g. chunked responses, or servers that hold the
        connection open while the caller is blocked on some external
        event). If the underlying socket is then closed/reset while the
        body is still being read, that failure surfaces here — not from
        the initial `fetch()` call above — so it must be wrapped the same
        way, otherwise callers can't distinguish it from a JSON parsing
        error and it won't be recognized as a connectivity failure.

apps/clear-signing-tester/src/infrastructure/state-handlers/OptOutStateHandler.ts
```diff
@@ -4,10 +4,17 @@ import { inject, injectable } from "inversify";
...
 import { type ScreenshotSaver } from "@root/src/domain/adapters/ScreenshotSaver";
+import { type CliLogLevel } from "@root/src/domain/models/config/LoggerConfig";
 import { type SignableInput } from "@root/src/domain/models/SignableInput";
+import { type RetryService } from "@root/src/domain/services/RetryService";
+import { type ScreenAnalyzerService } from "@root/src/domain/services/ScreenAnalyzer";
 
 import { type StateHandler, type StateHandlerResult } from "./StateHandler";
 
+const WEB3_CHECKS_MODAL_MAX_ATTEMPTS = 6;
+const WEB3_CHECKS_MODAL_POLL_DELAY_MS = 1000;
```
pvautherin-ledger, line 15: The interval between each Web3 Modal check attempts. Up to 6 × 1s = 6 s max wait.

apps/clear-signing-tester/src/infrastructure/state-handlers/OptOutStateHandler.ts
```diff
@@ -4,10 +4,17 @@ import { inject, injectable } from "inversify";
...
+const WEB3_CHECKS_MODAL_SETTLE_DELAY_MS = 1500;
```
pvautherin-ledger, line 16: Extra settle delay after the modal's text is first detected, before
actually tapping it.

The modal's text can be readable via the screen-content API
slightly before the Speculos automation layer is actually ready to accept
touch input on it.

.changeset/speculinho-signer-eth.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ledgerhq/device-signer-kit-ethereum": minor
```
pvautherin-ledger: Note for myself: to remove

.changeset/speculinho-context-module.md
```diff
@@ -0,0 +1,5 @@
+---
+"@ledgerhq/context-module": minor
```
pvautherin-ledger: Note for myself: to remove

packages/speculos-device-controller/src/internal/use-cases/touchUseCases.ts
```diff
@@ -90,3 +90,10 @@ export const acceptBlindSigning =
   <K extends string>(touch: TouchController<K>, deviceKey: K) =>
   async () =>
     await tapQuick(touch, deviceKey)({ x: 50, y: 94 });
+
+const WEB3_CHECKS_OPT_OUT_COORDS: PercentCoordinates = { x: 77, y: 94 };
```
pvautherin-ledger: consider using secondaryButton toucheEvent instead

PR #1665, opened 2026-07-22
🐛 (signer-eth): Fall back to legacy signing when the typed data flow throws

droplet-rl:
## 📝 Description

Thrown (non-`CommandResult`) errors in the typed-data flow bypass the `SignTypedDataLegacy` fallback and terminate the device action:

- `ProvideContext.onError` targeted the terminal `Error` state directly, while a *failed command result* from the same task routes through `ProvideContextResultCheck` → `notRefusedByUser` → `SignTypedDataLegacy`.
- `SignTypedData.onError` targeted `DetectBlindSigning` (error reporting) with no fallback attempt, while a failed command result routes through `SignTypedDataResultCheck` → `notRefusedByUser` → `SignTypedDataLegacy`.

The asymmetry matters most for **large EIP-712 payloads** (e.g. Safe multisig confirmations embedding tens of KB of calldata): providing the context streams the struct in many APDU chunks plus clear-signing context fetches, and failures on that path frequently surface as thrown transport/timeout errors rather than failed command results. The result is that oversized typed data is unsignable in DMK-based wallets, with no recovery, even though `SignTypedDataLegacyTask` handles exactly this case (Rabby ≥ v0.93.98 users hit this in production; report with reproduction details in the companion PR RabbyHub/Rabby#3930).

This PR routes both `onError` transitions through the same result-check states used for failed command results, so:

- thrown errors that are not a user refusal now fall back to legacy hash signing,
- user refusals (`0x6985`) keep their current behaviour (no fallback, error surfaces),
- errors during the fallback itself keep their current behaviour.

## ❓ Context

- No JIRA/issue — external contribution (`no-issue`). Happy to file a GitHub issue with full reproduction data if preferred.

## ✅ Checklist

- [x] Changeset added (`@ledgerhq/device-signer-kit-ethereum`, patch)
- [x] Tests updated/added:
  - `Error thrown while providing context should fallback to legacy signing` (updated from the previous terminal-error expectation)
  - `Error thrown while providing context with a user refusal should not fallback` (new — pins the 0x6985 guard)
  - `Error thrown while signing should fallback to legacy signing` (new)
- [x] `vitest` green for the package, lint/prettier pass on changed files

## 🧐 Checklist for the PR Reviewers

- The fallback semantics are unchanged for failed command results; this only makes thrown errors take the same route.
- `assignErrorFromEvent` stores the thrown error in `_internalState.error`, which is exactly what the `notRefusedByUser` guard inspects, so refusal detection behaves identically for both error shapes.

### DIFF
--- a/.changeset/typed-data-legacy-fallback-on-thrown-errors.md
+++ b/.changeset/typed-data-legacy-fallback-on-thrown-errors.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/device-signer-kit-ethereum": patch
+---
+
+Fall back to legacy hash signing when the typed data flow throws instead of returning a failed command result (e.g. transport errors while streaming large EIP-712 payloads). Previously, thrown errors during ProvideContext and SignTypedData bypassed the SignTypedDataLegacy fallback and terminated the device action, leaving oversized typed data unsignable. User refusals (0x6985) still surface as errors without fallback.
--- a/packages/signer/signer-eth/src/internal/app-binder/device-action/SignTypedData/SignTypedDataDeviceAction.test.ts
+++ b/packages/signer/signer-eth/src/internal/app-binder/device-action/SignTypedData/SignTypedDataDeviceAction.test.ts
@@ -932,7 +932,7 @@ describe("SignTypedDataDeviceAction", () => {
         });
       }));
 
-    it("Error thrown while providing context", () =>
+    it("Error thrown while providing context should fallback to legacy signing", () =>
       new Promise<void>((resolve, reject) => {
         setupOpenAppDAMock();
         setupAppConfig("1.15.0", false, false);
@@ -960,6 +960,15 @@ describe("SignTypedDataDeviceAction", () => {
         );
         buildContextMock.mockResolvedValueOnce(TEST_BUILT_CONTEXT);
         provideContextMock.mockRejectedValueOnce(new UnknownDAError("Error"));
+        signTypedDataLegacyMock.mockResolvedValueOnce(
+          CommandResultFactory({
+            data: {
+              v: 0x1c,
+              r: "0x8a540510e13b0f2b11a451275716d29e08caad07e89a1c84964782fb5e1ad788",
+              s: "0x64a0de235b270fbe81e8e40688f4a9f9ad9d283d690552c9331d7773ceafa513",
+            },
+          }),
+        );
 
         const expectedStates: Array<SignTypedDataDAState> = [
           {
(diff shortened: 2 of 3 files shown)


droplet-rl:
## How to verify

**Without a device:** the updated unit tests pin the exact behaviour change — `Error thrown while providing context should fallback to legacy signing`, `Error thrown while signing should fallback to legacy signing`, and (guard preservation) `Error thrown while providing context with a user refusal should not fallback`. `pnpm --filter @ledgerhq/device-signer-kit-ethereum test` → 603/603.

**With a device (sample app):** sign typed data whose `data` bytes field exceeds the Ethereum app's clear-signing capacity — ~16 KB+ on a Nano X. Generate a harmless payload in any console (`0x…dEaD` domain, so the signature is meaningless):

```js
const typedData = {
  types: {
    EIP712Domain: [
      { name: 'chainId', type: 'uint256' },
      { name: 'verifyingContract', type: 'address' },
    ],
    SafeTx: [
      { name: 'to', type: 'address' }, { name: 'value', type: 'uint256' },
      { name: 'data', type: 'bytes' }, { name: 'operation', type: 'uint8' },
      { name: 'safeTxGas', type: 'uint256' }, { name: 'baseGas', type: 'uint256' },
      { name: 'gasPrice', type: 'uint256' }, { name: 'gasToken', type: 'address' },
      { name: 'refundReceiver', type: 'address' }, { name: 'nonce', type: 'uint256' },
    ],
  },
  primaryType: 'SafeTx',
  domain: { chainId: 1, verifyingContract: '0x000000000000000000000000000000000000dEaD' },
  message: {
    to: '0x000000000000000000000000000000000000dEaD',
    value: '0',
    data: '0x' + 'ab'.repeat(20 * 1024), // 20 KB
    operation: 0, safeTxGas: '0', baseGas: '0', gasPrice: '0',
    gasToken: '0x0000000000000000000000000000000000000000',
    refundReceiver: '0x0000000000000000000000000000000000000000',
    nonce: '999999999',
  },
};
copy(JSON.stringify(typedData));
```

Paste into the sample app's *Sign typed message* input, Blind signing enabled on the device.

- **Before:** the device action terminates in error when streaming fails.
- **After:** the action falls through to `SignTypedDataLegacy` and the device prompts for the domain/message hashes.

Control: `1024` instead of `20 * 1024` → normal clear signing, both versions. On-device reject → error surfaces, no fallback, both versions.

This is the failure our signers hit in production with 15.8–38 KB Safe transactions (context in RabbyHub/Rabby#3930).

mbertin-ledger:
Hi @droplet-rl , 
Thx for your contribution, we are already aware of this issue, and this should be solved in the coming months ( around september/october) with a new Ethereum application release and the way we are interacting with the app for typed messages. 
I will circle back with the team, to see if it's useful or not to merge what you have done as we will surely refactor the `SignTypedDataDeviceAction` 
Letting you know.

PR #1683, opened 2026-07-27
✨ (signer-icp) [LIVE-34592]: Add update-call signing and neuron-stake flag

lysyi3m:
### 📝 Description

Extends `@ledgerhq/device-signer-kit-icp` with the two signing shapes ICP staking (neuron management) needs, on top of the transfer surface. Nothing existing changes behaviour; both additions are opt-in.

**1. Update-call signing (`signUpdateCall`)** — most neuron operations (dissolve, follow, split, hot-keys, maturity, …) are IC update calls to the NNS governance canister. Each must be signed together with its companion read-state request. The new capability takes the call body and the read-state body as **separate inputs**, drives the device's combined-sign path (`INS 0x03`, chunked), and returns **two signatures** (request + read-state) plus the read-state body, so the caller can submit the call and poll its status while keeping the request ↔ read-state pairing intact.

**2. Neuron-creation flag on `signTransaction`** — neuron creation is a governance-subaccount transfer, the same device sign instruction as a plain transfer (`INS 0x02`) distinguished by a transaction-type flag. `signTransaction` gains an optional `stake` flag that sets P2 accordingly.

Implementation follows the existing kit structure and the `signer-eth` precedents: the same-INS P2 sub-mode stays a parameter on `SignTransactionCommand`; the different-INS combined-sign is a new `SignUpdateCallCommand` with its own response parser (the `0x03` response is a fixed 192-byte block — two 32-byte digests each with a 64-byte r‖s signature, no `v`/DER — distinct from the `0x02` layout). Wire format and response layout were verified against the device app source (`LedgerHQ/app-icp`) and the reference host library.

```ts
// neuron management
const { requestSignature, readStateSignature, readStateBody } =
  await signer.signUpdateCall(path, callRequest, readStateRequest);
// neuron creation
await signer.signTransaction(path, transferBlob, { stake: true });
```

### ❓ Context

- **JIRA or GitHub link**: [LIVE-34592](https://ledgerhq.atlassian.net/browse/LIVE-34592)

### ✅ Checklist

- [x] **Covered by automatic tests** — unit tests for the new command (response split incl. per-field golden fixture, chunk boundaries, malformed/short/long rejection, refusal status word), the task (combined-message framing vector asserting read-state-first ordering and LE lengths, chunk boundaries, missing-body rejection, error paths), use-case, app-binder, DI binding, and the stake flag across command/task/use-case. 104 tests, 99.3% line coverage.
- [x] **Changeset is provided** — `device-signer-kit-icp` minor (additive; consistent with the `signer-eth` EIP-7702 method addition shipped as minor).
- [x] **Documentation is up-to-date** — public API surfaced through the `SignerIcp` interface and barrel exports.
  - New `signUpdateCall` capability (`INS 0x03`) — no effect on existing transfer/address/version flows.
  - New optional `stake` flag on `signTransaction` — defaults to off (P2 unchanged).
  - QA focus: on-device update-call signing for a neuron-management operation, and a neuron-creation transfer.

[LIVE-34592]: https://ledgerhq.atlassian.net/browse/LIVE-34592?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/signer/signer-icp/src/api/SignerIcp.ts
+++ b/packages/signer/signer-icp/src/api/SignerIcp.ts
@@ -1,14 +1,20 @@
 import { type GetAddressDAReturnType } from "@api/app-binder/GetAddressDeviceActionTypes";
 import { type GetVersionDAReturnType } from "@api/app-binder/GetVersionDeviceActionTypes";
 import { type SignTransactionDAReturnType } from "@api/app-binder/SignTransactionDeviceActionTypes";
+import { type SignUpdateCallDAReturnType } from "@api/app-binder/SignUpdateCallDeviceActionTypes";
 
 export type AddressOptions = {
   checkOnDevice?: boolean;
   skipOpenApp?: boolean;
 };
 
-export type TransactionOptions = {
+export type CommonOptions = {
   skipOpenApp?: boolean;
+};
+
+export type TransactionOptions = CommonOptions & {
+  // Sign a neuron-creation transfer (governance-subaccount) instead of a plain transfer.
+  stake?: boolean;
 };
 
 export interface SignerIcp {
@@ -24,4 +30,13 @@ export interface SignerIcp {
     transaction: Uint8Array,
     options?: TransactionOptions,
   ) => SignTransactionDAReturnType;
+
+  // Sign an IC update call together with its companion read-state request
+  // (neuron management). Returns both signatures with the read-state body.
+  signUpdateCall: (
+    derivationPath: string,
+    callRequest: Uint8Array,
+    readStateRequest: Uint8Array,
+    options?: CommonOptions,
+  ) => SignUpdateCallDAReturnType;
 }
--- a/packages/signer/signer-icp/src/api/app-binder/SignUpdateCallDeviceActionTypes.ts
+++ b/packages/signer/signer-icp/src/api/app-binder/SignUpdateCallDeviceActionTypes.ts
@@ -1,0 +1,30 @@
+import {
+  type CallTaskInAppDAOutput,
+  type CommandErrorResult,
+  type ExecuteDeviceActionReturnType,
+  type OpenAppDAError,
+  type OpenAppDARequiredInteraction,
+  type UserInteractionRequired,
+} from "@ledgerhq/device-management-kit";
+
+import { type UpdateCallSignature } from "@api/model/UpdateCallSignature";
+import { type IcpErrorCodes } from "@internal/app-binder/command/utils/IcpApplicationErrors";
+
+export type SignUpdateCallDAOutput = CallTaskInAppDAOutput<UpdateCallSignature>;
+export type SignUpdateCallDAError =
+  | OpenAppDAError
+  | CommandErrorResult<IcpErrorCodes>["error"];
+
+type SignUpdateCallDARequiredInteraction =
+  | OpenAppDARequiredInteraction
+  | UserInteractionRequired.SignTransaction;
+
+export type SignUpdateCallDAIntermediateValue = {
+  requiredUserInteraction: SignUpdateCallDARequiredInteraction;
+};
+
+export type SignUpdateCallDAReturnType = ExecuteDeviceActionReturnType<
+  SignUpdateCallDAOutput,
+  SignUpdateCallDAError,
+  SignUpdateCallDAIntermediateValue
+>;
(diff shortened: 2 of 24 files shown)


packages/signer/signer-icp/src/api/SignerIcp.ts
```diff
@@ -24,4 +27,13 @@ export interface SignerIcp {
...
     options?: TransactionOptions,
   ) => SignTransactionDAReturnType;
+
+  // Sign an IC update call together with its companion read-state request
+  // (neuron management). Returns both signatures with the read-state body.
+  signUpdateCall: (
+    derivationPath: string,
+    callRequest: Uint8Array,
+    readStateRequest: Uint8Array,
+    options?: TransactionOptions,
```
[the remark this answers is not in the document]
lysyi3m: Good catch — fixed. `signUpdateCall` now takes a dedicated `CommonOptions` (`{ skipOpenApp? }`) instead of `TransactionOptions`, so the transfer-only `stake` flag can no longer be passed here. `TransactionOptions` is now `CommonOptions & { stake? }`.

packages/signer/signer-icp/src/internal/app-binder/command/SignUpdateCallCommand.ts
```diff
@@ -0,0 +1,186 @@
...
+export const P2_UPDATE_CALL = 0x00;
+
+const DERIVATION_PATH_LENGTH = 5;
+// The last-chunk response is a fixed 192-byte block: for the request and the
+// read-state each, a 32-byte signed digest followed by a 64-byte r‖s signature.
+const HASH_LENGTH = 32;
+const SIGNATURE_R_LENGTH = 32;
+const SIGNATURE_S_LENGTH = 32;
+
+export enum SignPhase {
```
fAnselmi-Ledger: [SHOULD] this enum was also declared in `signTransactionCommand.ts`, should be shared from a constant file!

packages/signer/signer-icp/src/internal/app-binder/IcpAppBinder.ts
```diff
@@ -84,4 +87,29 @@ export class IcpAppBinder {
...
+            new SignUpdateCallTask(
+              internalApi,
+              args,
+              this.dmkLoggerFactory("SignUpdateCallTask"),
+            ).run(),
+          appName: APP_NAME,
+          requiredUserInteraction: UserInteractionRequired.SignTransaction,
+          skipOpenApp: args.skipOpenApp ?? false,
+        },
+        logger: this.dmkLoggerFactory("SignUpdateCallCommand"),
```
fAnselmi-Ledger: [SHOULD] should be called `SignUpdateCallTask`

packages/signer/signer-icp/src/api/app-binder/SignUpdateCallDeviceActionTypes.ts
```diff
@@ -0,0 +1,31 @@
...
+  type OpenAppDARequiredInteraction,
+  type SendCommandInAppDAOutput,
+  type UserInteractionRequired,
+} from "@ledgerhq/device-management-kit";
+
+import { type UpdateCallSignature } from "@api/model/UpdateCallSignature";
+import { type IcpErrorCodes } from "@internal/app-binder/command/utils/IcpApplicationErrors";
+
+export type SignUpdateCallDAOutput =
+  SendCommandInAppDAOutput<UpdateCallSignature>;
```
fAnselmi-Ledger: [SHOULD] we have a `CallTaskInAppDAOutput<T>` for this!

packages/signer/signer-icp/src/internal/app-binder/command/SignUpdateCallCommand.ts
lysyi3m: Done — moved `SignPhase` and `P1_INIT/ADD/LAST` into `app-binder/constants.ts` and import them directly in both commands (and the tasks/tests), matching how signer-zcash shares its APDU header constants. Left the P2 values per-command since they differ (stake flag vs update-call).

packages/signer/signer-icp/src/internal/app-binder/IcpAppBinder.ts
lysyi3m: Done — renamed the tag to `SignUpdateCallTask`.

packages/signer/signer-icp/src/api/app-binder/SignUpdateCallDeviceActionTypes.ts
lysyi3m: Good call — switched to `CallTaskInAppDAOutput<UpdateCallSignature>`, since it's driven by a `CallTaskInAppDeviceAction`.

fAnselmi-Ledger (approved):

PR #1687, opened 2026-07-30
⚡ (ci) [NO-ISSUE]: Consolidate checks into single Turbo job with --affected

aussedatlo:
### 📝 Description

Speeds up PR / merge-queue CI by restructuring how checks run, and unblocks Turbo's `--affected` so it can actually scope work per PR.

**1. Consolidate all checks into a single Turbo job (removes redundant cache I/O).**
Previously `build-libraries`, `prettier`, `lint`, `typecheck`, `tests`, and `sample-integration` were separate jobs that each re-ran setup + `pnpm install` + a per-job `**/lib` build-cache upload/download keyed by SHA. That cache transfer (paid ~6×) was the dominant cost (~8 min), **not** build compute — a cold build is ~3s (esbuild). Now everything runs as sequential steps in one `checks` job, so build output stays on disk between steps and the cross-job cache transfer is gone entirely.

- Each task runs `turbo run <task> --affected` (base = PR base branch / merge-group base sha).
- Check steps use `continue-on-error` so one failure doesn't hide the others; a final **Report** step writes a ✅/❌ per-step table to the job summary and gates the job.
- **Playwright (sample E2E) is folded in** as steps of the same job — it reuses the libs already built instead of re-running setup/build on a second runner — and is **gated on `--affected`**: the browser install + test run are skipped unless `@ledgerhq/device-management-kit-sample` (or one of its deps) is in the affected set.
- `cs-tester` / `solana-cs-tester` stay separate (Speculos runners, already gated by `detect-changes`); `health-check` remains the required-status aggregator (now `needs: [checks]`).
- Same job consolidation applied to `merge_queue.yml` (which has no Playwright step).

**2. Remove the root `package.json` workspace deps so `--affected` scopes correctly.**
The root depended on `@ledgerhq/ldmk-cli` and `@ledgerhq/ldmk-tool` (`workspace:^`). `ldmk-cli` transitively pulls `dmk`, `signer-eth`, `context-module`, etc. into the root's dependency closure, and Turbo treats the root package as an implicit dependency of every workspace package — so a change to any of those core packages marked **all** packages affected, defeating `--affected`. Dropping both root deps (the `cli` script already uses `--filter`; the `tool` script is converted to `pnpm --filter @ledgerhq/ldmk-tool exec node cli.cjs`) fixes it.

Measured effect of the `--affected` fix (steady state, i.e. once merged to develop):

| Change | Before | After |
| --- | --- | --- |
| signer-eth | 46 (all) | **4** |
| context-module | 46 (all) | **10** |
| dmk | 46 (all) | 37 (dmk is a genuine universal dep) |

> Note: the PR that removes the root deps itself touches `pnpm-lock.yaml` (a Turbo global trigger), so this PR's own run still sees all packages affected. The benefit applies to every PR after it lands on develop.

**Why sequential (not parallel lint/typecheck/test)?** Benchmarked both cold: for a whole-monorepo change, sequential quality phase = 108s vs 206s combined — running `vitest`/`tsc`/`eslint` concurrently oversubscribes the CPU and thrashes, since each already saturates all cores. Combined only wins on tiny scoped sets (41s vs 59s) where absolute time is already low. Sequential wins the case that costs the most, and gives the clean per-step report.

### ❓ Context

- **Feature**: CI build/cache speedup + `--affected` scoping

### ✅ Checklist

- [ ] **Covered by automatic tests** — N/A, CI workflow + root tooling change; validated by this PR's own run.
- [ ] **Changeset is provided** — No changeset needed; changes only affect CI workflows and the private root package (`.github/**`, root `package.json`), no published package.
- [ ] **Documentation is up-to-date** — N/A
  - CI structure changes; reviewers/QA should confirm the `checks` job (build, prettier, lint, typecheck, test, Playwright) and the cs-tester jobs all run, and that the report step surfaces failures.
  - Playwright now runs inside `checks` and is skipped when the sample app is unaffected — confirm it still runs when the sample app/deps change.
  - `pnpm cli` and `pnpm tool` still work (root workspace deps removed; `tool` now runs via `--filter … exec`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/merge_queue.yml
+++ b/.github/workflows/merge_queue.yml
@@ -10,71 +10,100 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  build-libraries:
-    name: Build libraries
+  checks:
+    name: Checks (build, prettier, lint, typecheck, test)
+    # Single job running each Turbo task as its own step. Build output stays on
+    # disk between steps (no cross-job build-cache upload/download) and
+    # `--affected` restricts every task to the packages changed in the group.
     runs-on: "ledgerhq-device-sdk"
+    env:
+      # Base commit used by `turbo run --affected` to detect changed packages.
+      TURBO_SCM_BASE: ${{ github.event.merge_group.base_sha }}
     steps:
       - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
+        with:
+          # Full history is required so Turbo can compute the merge-base for
+          # `--affected`.
+          fetch-depth: 0
 
-      - uses: LedgerHQ/device-sdk-ts/.github/actions/setup-with-build-cache-composite@b83ec5020b992dcc53f683830e0abd2e147791ae
+      - uses: LedgerHQ/device-sdk-ts/.github/actions/setup-with-cache-composite@b83ec5020b992dcc53f683830e0abd2e147791ae
 
-  prettier:
-    name: Run prettier
-    needs: [build-libraries]
-    runs-on: "ledgerhq-device-sdk"
-    steps:
-      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7
-
-      - uses: LedgerHQ/device-sdk-ts/.github/actions/setup-with-build-cache-composite@b83ec5020b992dcc53f683830e0abd2e147791ae
+      - name: Build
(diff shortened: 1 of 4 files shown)


aussedatlo:
Superseded by #1692 (moved to branch chore/dsdk-1402-ci-turbo-affected, tracked under DSDK-1402).

PR #1466, opened 2026-05-04
✨ (signer-zcash) [LIVE-23291]: Add full viewing key export

may01:
### 📝 Description

This change adds full viewing key (FVK) export for Zcash on the device signer: new public API, device flow (command / task / use case), error and APDU helpers where needed, and sample app wiring so the flow can be exercised manually.

Why: Wallets and integrations need the FVK to watch balances and build transactions without holding spend authority; the key must come from the device when policy requires it.

### ❓ Context

- **JIRA or GitHub link**: [LIVE-23291](https://ledgerhq.atlassian.net/browse/LIVE-23291)
- Device part was implemented in [NAPPS-1261](https://ledgerhq.atlassian.net/browse/NAPPS-1261)

- **Feature**: Export Zcash full viewing key from the Ledger Zcash app via @ledgerhq/device-signer-kit-zcash, with tests and sample UI.

### ✅ Checklist

Pull Requests must pass CI checks and undergo code review. Set the PR as Draft if it is not yet ready for review.

- [x] **Covered by automatic tests** 
- [x] **Changeset is provided** 
- [x] **Documentation is up-to-date** 
- [x] **Impact of the changes:** 
  - introduces full viewing key export functionality for zcash with, commands, tasks, and binder update

---

### 🧐 Checklist for the PR Reviewers

- [ ] **The code aligns with the requirements** described in the linked JIRA or GitHub issue.
- [ ] **The PR description clearly documents the changes** made and explains any technical trade-offs or design decisions.
- [ ] **There are no undocumented trade-offs**, technical debt, or maintainability issues.
- [ ] **The PR has been tested** thoroughly, and any potential edge cases have been considered and handled.
- [ ] **Any new dependencies** have been justified and documented.

[LIVE-23291]: https://ledgerhq.atlassian.net/browse/LIVE-23291?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[NAPPS-1261]: https://ledgerhq.atlassian.net/browse/NAPPS-1261?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1695, opened 2026-07-30
✨ (signer-zcash) [LIVE-35012]: Add getShieldedAddress

vladyslavchupovskiy-ext-art:
### 📝 Description

Adds `getShieldedAddress` to `SignerZcash`, wrapping INS_GET_SHIELD_ADDR (0x51) from app-zcash ≥ v3.8.0.

The APDU sends two length-prefixed derivation paths to the device — Orchard (`32'/coin/account'`, 3 hardened levels) and transparent (`44'/coin/account'/change/index`, 5 levels) — and returns a unified address (`u1…`) with a single Orchard receiver. The device can optionally display the address for user confirmation (`checkOnDevice`).

**Usage:**
```ts
const result = signer.getShieldedAddress("44'/133'/0'/0/0", { checkOnDevice: true });
The public API takes a single transparent derivation path; the task layer derives the Orchard account path from it, ensuring account indexes always match.
```
❓ Context

- JIRA or GitHub link: LIVE-35012

✅ Checklist

- [x] Covered by automatic tests — command (getApdu + parseResponse), task (path derivation, command forwarding), and use case layers all have unit tests; 192 tests passing
- [x] Changeset is provided — minor bump for @ledgerhq/device-signer-kit-zcash
- [x] Documentation is up-to-date — README.md and apps/docs/content/docs/references/signers/zcash.mdx updated with usage, parameters, and return type
- [x] Impact of the changes:
  - Adds getShieldedAddress to SignerZcash interface and DefaultSignerZcash — additive only, no existing API changed
  - New files: GetShieldedAddressCommand, GetShieldedAddressTask, GetShieldedAddressUseCase, GetShieldedAddressDeviceActionTypes
  - DI: new symbol and binding in addressTypes / addressModule
  - Requires app-zcash ≥ v3.8.0 (two-path form of 0x51); v3.5.0 single-path form is not used
  - QA should verify: address displayed on device matches host-side derivation, checkOnDevice: false skips device display, 6985 (user rejection) surfaces as typed error

### DIFF
--- a/.changeset/signer-zcash-get-shielded-address.md
+++ b/.changeset/signer-zcash-get-shielded-address.md
@@ -1,0 +1,7 @@
+---
+"@ledgerhq/device-signer-kit-zcash": minor
+---
+
+Add `getShieldedAddress` to `SignerZcash` for INS_GET_SHIELD_ADDR (0x51).
+
+Sends the transparent derivation path (`44'/coin/account'/change/index`) together with the derived Orchard account path (`32'/coin/account'`) to the device. The device uses the transparent path for account matching and on-device display; the returned unified address contains a single Orchard receiver. Pass `checkOnDevice: true` to have the device display the address for user confirmation.
--- a/packages/signer/signer-zcash/src/api/SignerZcash.ts
+++ b/packages/signer/signer-zcash/src/api/SignerZcash.ts
@@ -1,6 +1,7 @@
 import { type GetAddressDAReturnType } from "@api/app-binder/GetAddressDeviceActionTypes";
 import { type GetAppConfigDAReturnType } from "@api/app-binder/GetAppConfigDeviceActionTypes";
 import { type GetFullViewingKeyDAReturnType } from "@api/app-binder/GetFullViewingKeyDeviceActionTypes";
+import { type GetShieldedAddressDAReturnType } from "@api/app-binder/GetShieldedAddressDeviceActionTypes";
 import { type GetTrustedInputDAReturnType } from "@api/app-binder/GetTrustedInputActionTypes";
 import { type SignMessageDAReturnType } from "@api/app-binder/SignMessageDeviceActionTypes";
 import { type SignPcztTransactionDAReturnType } from "@api/app-binder/SignPcztTransactionDeviceActionTypes";
@@ -23,6 +24,19 @@ export interface SignerZcash {
     derivationPath: string,
     options?: FullViewingKeyOptions,
   ) => GetFullViewingKeyDAReturnType;
+
+  /**
+   * Returns a unified shielded address (single Orchard receiver) for the given
+   * transparent derivation path. `derivationPath` must be a 5-level BIP44 path
+   * (`44'/coin'/account'/change/index`); the Orchard account path
+   * (`32'/coin'/account'`) is derived from it automatically by the task.
+   * Pass `checkOnDevice: true` to have the address displayed on the device for
+   * user confirmation.
+   */
+  getShieldedAddress: (
+    derivationPath: string,
+    options?: AddressOptions,
+  ) => GetShieldedAddressDAReturnType;
 
   signTransaction: (
     args: LegacyCreateTransactionArg,
--- a/packages/signer/signer-zcash/src/internal/app-binder/ZcashAppBinder.ts
+++ b/packages/signer/signer-zcash/src/internal/app-binder/ZcashAppBinder.ts
@@ -11,6 +11,7 @@ import { inject, injectable } from "inversify";
 import { type GetAddressDAReturnType } from "@api/app-binder/GetAddressDeviceActionTypes";
 import { type GetAppConfigDAReturnType } from "@api/app-binder/GetAppConfigDeviceActionTypes";
 import { type GetFullViewingKeyDAReturnType } from "@api/app-binder/GetFullViewingKeyDeviceActionTypes";
+import { type GetShieldedAddressDAReturnType } from "@api/app-binder/GetShieldedAddressDeviceActionTypes";
 import { type GetTrustedInputDAReturnType } from "@api/app-binder/GetTrustedInputActionTypes";
 import { type SignMessageDAReturnType } from "@api/app-binder/SignMessageDeviceActionTypes";
 import { type SignPcztTransactionDAReturnType } from "@api/app-binder/SignPcztTransactionDeviceActionTypes";
@@ -25,6 +26,7 @@ import { GetAddressCommand } from "./command/GetAddressCommand";
 import { GetAppConfigCommand } from "./command/GetAppConfigCommand";
 import { SignMessageCommand } from "./command/SignMessageCommand";
 import { GetFullViewingKeyTask } from "./task/GetFullViewingKeyTask";
+import { GetShieldedAddressTask } from "./task/GetShieldedAddressTask";
 import { GetTrustedInputTask } from "./task/GetTrustedInputTask";
 import { SignPcztTransactionTask } from "./task/SignPcztTransactionTask";
 import { SignTransactionTask } from "./task/SignTransactionTask";
@@ -60,6 +62,30 @@ export class ZcashAppBinder {
       deviceAction: new SendCommandInAppDeviceAction({
         input: {
           command: new GetAddressCommand(args),
+          appName: APP_NAME,
+          requiredUserInteraction: args.checkOnDevice
+            ? UserInteractionRequired.VerifyAddress
+            : UserInteractionRequired.None,
+          skipOpenApp: args.skipOpenApp,
+        },
+      }),
+    });
+  }
+
+  getShieldedAddress(args: {
+    derivationPath: string;
+    checkOnDevice: boolean;
+    skipOpenApp: boolean;
+  }): GetShieldedAddressDAReturnType {
+    return this.dmk.executeDeviceAction({
+      sessionId: this.sessionId,
+      deviceAction: new CallTaskInAppDeviceAction({
+        input: {
+          task: async (internalApi: InternalApi) =>
+            new GetShieldedAddressTask(internalApi, {
+              derivationPath: args.derivationPath,
+              checkOnDevice: args.checkOnDevice,
+            }).run(),
           appName: APP_NAME,
           requiredUserInteraction: args.checkOnDevice
             ? UserInteractionRequired.VerifyAddress
--- a/packages/signer/signer-zcash/src/internal/app-binder/command/GetShieldedAddressCommand.test.ts
+++ b/packages/signer/signer-zcash/src/internal/app-binder/command/GetShieldedAddressCommand.test.ts
@@ -1,0 +1,318 @@
+import {
+  ApduResponse,
+  isSuccessCommandResult,
+} from "@ledgerhq/device-management-kit";
+
+import { GetShieldedAddressCommand } from "@internal/app-binder/command/GetShieldedAddressCommand";
+import { ZCASH_CLA } from "@internal/app-binder/command/utils/apduHeaderUtils";
+import { type ZcashAppCommandError } from "@internal/app-binder/command/utils/zcashApplicationErrors";
+
+const GET_SHIELDED_ADDRESS_INS = 0x51;
+
+// 32'/133'/0' + 44'/133'/0'/0/0 — no display
+const GET_SHIELDED_ADDRESS_APDU_NO_DISPLAY = Uint8Array.from([
+  ZCASH_CLA,
+  GET_SHIELDED_ADDRESS_INS,
+  0x00, // P1 (no display)
+  0x00, // P2
+  0x22, // Lc: 1+12+1+20 = 34
+  // Orchard path: 32'/133'/0'
+  0x03,
+  0x80,
+  0x00,
+  0x00,
+  0x20, // 32'
+  0x80,
+  0x00,
+  0x00,
+  0x85, // 133'
+  0x80,
+  0x00,
+  0x00,
+  0x00, // 0'
+  // Transparent path: 44'/133'/0'/0/0
+  0x05,
+  0x80,
+  0x00,
+  0x00,
+  0x2c, // 44'
+  0x80,
+  0x00,
+  0x00,
+  0x85, // 133'
+  0x80,
+  0x00,
+  0x00,
+  0x00, // 0'
+  0x00,
+  0x00,
+  0x00,
+  0x00, // 0
+  0x00,
+  0x00,
+  0x00,
+  0x00, // 0
+]);
+
+// same paths — display on device
+const GET_SHIELDED_ADDRESS_APDU_WITH_DISPLAY = Uint8Array.from([
+  ZCASH_CLA,
+  GET_SHIELDED_ADDRESS_INS,
+  0x01, // P1 (display)
+  0x00,
+  0x22,
+  0x03,
+  0x80,
+  0x00,
+  0x00,
+  0x20,
+  0x80,
+  0x00,
+  0x00,
+  0x85,
+  0x80,
+  0x00,
+  0x00,
+  0x00,
+  0x05,
+  0x80,
+  0x00,
+  0x00,
+  0x2c,
+  0x80,
+  0x00,
+  0x00,
+  0x85,
+  0x80,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+]);
+
+// 32'/133'/1' + 44'/133'/1'/0/0 — account 1, no display
+const GET_SHIELDED_ADDRESS_APDU_ACCOUNT_1 = Uint8Array.from([
+  ZCASH_CLA,
+  GET_SHIELDED_ADDRESS_INS,
+  0x00,
+  0x00,
+  0x22,
+  0x03,
+  0x80,
+  0x00,
+  0x00,
+  0x20,
+  0x80,
+  0x00,
+  0x00,
+  0x85,
+  0x80,
+  0x00,
+  0x00,
+  0x01, // 1'
+  0x05,
+  0x80,
+  0x00,
+  0x00,
+  0x2c,
+  0x80,
+  0x00,
+  0x00,
+  0x85,
+  0x80,
+  0x00,
+  0x00,
+  0x01, // 1'
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+]);
+
+// 32'/133'/0' + 44'/133'/0'/0/5 — proves address index is serialized
+const GET_SHIELDED_ADDRESS_APDU_ADDR_INDEX_5 = Uint8Array.from([
+  ZCASH_CLA,
+  GET_SHIELDED_ADDRESS_INS,
+  0x00,
+  0x00,
+  0x22,
+  0x03,
+  0x80,
+  0x00,
+  0x00,
+  0x20,
+  0x80,
+  0x00,
+  0x00,
+  0x85,
+  0x80,
+  0x00,
+  0x00,
+  0x00,
+  0x05,
+  0x80,
+  0x00,
+  0x00,
+  0x2c,
+  0x80,
+  0x00,
+  0x00,
+  0x85,
+  0x80,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x00,
+  0x05, // address index = 5
+]);
+
+function buildSuccessResponse(address: string): ApduResponse {
+  const addressBytes = new TextEncoder().encode(address);
+  const data = new Uint8Array(2 + addressBytes.length);
+
+  // u16 BE length prefix
+  data[0] = (addressBytes.length >> 8) & 0xff;
+  data[1] = addressBytes.length & 0xff;
+
+  // UTF-8 address
+  data.set(addressBytes, 2);
+
+  return new ApduResponse({
+    statusCode: new Uint8Array([0x90, 0x00]),
+    data,
+  });
+}
+
+describe("GetShieldedAddressCommand", () => {
+  const defaultArgs = {
+    transparentDerivationPath: "44'/133'/0'/0/0",
+    orchardDerivationPath: "32'/133'/0'",
+    checkOnDevice: false,
+  };
+
+  describe("name", () => {
+    it("should be 'GetShieldedAddress'", () => {
+      const command = new GetShieldedAddressCommand(defaultArgs);
+      expect(command.name).toBe("GetShieldedAddress");
+    });
+  });
+
+  describe("getApdu", () => {
+    it("should return correct APDU with checkOnDevice false", () => {
+      const command = new GetShieldedAddressCommand(defaultArgs);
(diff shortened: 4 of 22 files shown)


packages/signer/signer-zcash/src/internal/app-binder/task/GetShieldedAddressTask.ts
```diff
@@ -0,0 +1,70 @@
...
+};
+
+export type GetShieldedAddressTaskResult = DmkResult<
+  { readonly address: string },
+  CommandErrorResult<ZcashErrorCodes>["error"]
+>;
+
+function deriveShieldedAddressPaths(
+  derivationPath: string,
+): { orchardDerivationPath: string; transparentDerivationPath: string } | { error: InvalidStatusWordError } {
```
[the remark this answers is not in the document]
vladyslavchupovskiy-ext-art: Done

packages/signer/signer-zcash/src/internal/app-binder/command/GetShieldedAddressCommand.test.ts
```diff
@@ -0,0 +1,219 @@
...
+}
+
+describe("GetShieldedAddressCommand", () => {
+
+  const defaultArgs = {
+    transparentDerivationPath: "44'/133'/0'/0/0",
+    orchardDerivationPath: "32'/133'/0'",
+    checkOnDevice: false
+  };
+
```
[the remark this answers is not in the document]
vladyslavchupovskiy-ext-art: Done

packages/signer/signer-zcash/src/internal/app-binder/ZcashAppBinder.ts
```diff
@@ -28,6 +28,8 @@ import { GetFullViewingKeyTask } from "./task/GetFullViewingKeyTask";
 import { GetTrustedInputTask } from "./task/GetTrustedInputTask";
 import { SignPcztTransactionTask } from "./task/SignPcztTransactionTask";
 import { SignTransactionTask } from "./task/SignTransactionTask";
+import { type GetShieldedAddressDAReturnType } from "@api/app-binder/GetShieldedAddressDeviceActionTypes";
+import { GetShieldedAddressTask } from "./task/GetShieldedAddressTask";
```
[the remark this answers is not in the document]
vladyslavchupovskiy-ext-art: Done

.changeset/signer-zcash-get-shielded-address.md
```diff
@@ -0,0 +1,7 @@
+---
+"@ledgerhq/signer-zcash": minor
```
cted-ledger: **Blocker** — this package does not exist in the workspace. The published name is `@ledgerhq/device-signer-kit-zcash` (see `packages/signer/signer-zcash/package.json`); `@ledgerhq/signer-zcash` is only the TypeScript path alias.

As written, this changeset yields no version bump and no CHANGELOG entry for the one package this PR actually modifies.

Danger's `checkChangesetSinglePackage` only counts how many packages a changeset declares (`releases.length > 1`) — it never checks the name against the workspace, which is why CI is green here. Running `pnpm changeset`, as `CONTRIBUTING.md` step 6 prescribes, picks the package from a generated list and makes this typo impossible.

cted-ledger (commented):
Two blockers inline — the changeset names a package that doesn't exist, and the only nominal-path test asserts nothing. Six non-blockers alongside them.

Three points have no line to anchor to:

**Docs not updated.** `packages/signer/signer-zcash/README.md` (numbered index of the 7 use cases) and `apps/docs/content/docs/references/signers/zcash.mdx` both omit `getShieldedAddress`. The "Documentation is up-to-date" box is unchecked with "no public docs affected".

**Test suites of the modified components not extended.** `ZcashAppBinder.ts`, `DefaultSignerZcash.ts` and `addressModule.ts` are modified; their existing suites cover every sibling method and none was updated. PR #1466 (`getFullViewingKey`, same package) extended all three.

**PR description.** The linked key is LIVE-33571, a merged hyperliquid ADR (#1666) — the task here is LIVE-35012. And the UA carries a single Orchard receiver, not both: `get_shielded_addr.rs` builds it with `UnifiedAddress::try_from_items(vec![Receiver::Orchard(..)])`; the transparent path is sent for account matching and on-device display.

packages/signer/signer-zcash/src/internal/app-binder/command/GetShieldedAddressCommand.test.ts
```diff
@@ -0,0 +1,316 @@
...
+  describe("parseResponse", () => {
+    const expectedAddress =
+      "u17qxnge3fpth2w43cfvz3lezxkevzmh5lpl5j4vlkfclpxdz9rx2fmml98wmwq3268yld6exrhyg29k2xhrnt4rldxva96qe8uwf7qc9";
+
+    const response = buildSuccessResponse(expectedAddress);
+    it("should return the unified address on success", () => {
+      const command = new GetShieldedAddressCommand(defaultArgs);
+      const res = command.parseResponse(response);
+
+      if (isSuccessCommandResult(res)) {
```
cted-ledger, line 274: **Blocker** — this assertion is vacuous. If `parseResponse` returns an error, `isSuccessCommandResult(res)` is `false`, the block is skipped, and the test passes green having asserted nothing. Since this is the only test covering the nominal path, a regression that breaks address parsing outright would go unnoticed.

The three tests just below (lines 286, 301, 313) get this right — they assert the predicate before narrowing on it. Same shape needed here:

```ts
expect(isSuccessCommandResult(res)).toBe(true);
if (isSuccessCommandResult(res)) {
  expect(res.data.address).toEqual(expectedAddress);
}
```

packages/signer/signer-zcash/src/internal/app-binder/command/GetShieldedAddressCommand.test.ts
```diff
@@ -0,0 +1,316 @@
...
+      const apdu = command.getApdu();
+      expect(apdu.getRawApdu()).toStrictEqual(
+        GET_SHIELDED_ADDRESS_APDU_ADDR_INDEX_5,
+      );
+    });
+  });
+
+  describe("parseResponse", () => {
+    const expectedAddress =
+      "u17qxnge3fpth2w43cfvz3lezxkevzmh5lpl5j4vlkfclpxdz9rx2fmml98wmwq3268yld6exrhyg29k2xhrnt4rldxva96qe8uwf7qc9";
```
cted-ledger: This unified address is invented. `app-zcash` publishes real device outputs for exactly the two path pairs used in this file, in `tests/standalone/test_pubkey_cmd.py`:

- `m/32'/133'/0'` + `m/44'/133'/0'/0/0` → `u1u2h4ce7e2cn3z4nzur95muq2dl4da9x8h8kdp2l80gm9nl9raj8zzpx79ycjnfvar4v5exea5pqr5y9qsnlp0cdunwf9yjjx5c4q7ar9`
- `m/32'/133'/1'` + `m/44'/133'/1'/0/0` → `u1n4d94z4l9zs0kxhhytwyktg3rsmr9u0eagt3kn78j9m3lmnuzswuwn63az5jzfwqmvrfn0g8s3rvvg0wr0pklnkejm6d69hv8u5g6w9e`

To be clear about what this changes: because `buildSuccessResponse` re-encodes whatever string it is given, the test stays a round-trip of its own encoder either way. The gain is traceability — the fixture becomes checkable against the firmware, and it exercises the real UA length rather than an arbitrary one. Worth taking since `defaultArgs` already are the `account-0` paths.

packages/signer/signer-zcash/src/internal/app-binder/task/GetShieldedAddressTask.ts
```diff
@@ -0,0 +1,72 @@
...
+  if (components.length !== 5) {
+    return {
+      error: new InvalidStatusWordError(
+        `Shielded address derivation path must have exactly 5 levels purpose/coin/account/change/index (got "${derivationPath}")`,
+      ),
+    };
+  }
+  const [, coin, account] = components;
+  return {
+    orchardDerivationPath: `32'/${coin}/${account}`,
```
cted-ledger: The Orchard path is rebuilt by string interpolation from components that were never validated — only their *count* is checked. Two consequences:

- `44'/133'/0/0/0` (unhardened account) produces `32'/133'/0`, which the device rejects: `check_bip44_compliance` in ZIP32 mode requires a hardened account (`app-zcash/src/utils.rs`).
- `0'/133'/0'/0/0` passes here and is only caught by the device, since `BIP44_ALLOWED_PURPOSES` is `[44, 32]`.

In both cases the user pays a full device round-trip to receive an opaque `6A80`. Validating purpose, coin type and account hardening host-side would fail fast with a message that names the actual problem.

packages/signer/signer-zcash/src/internal/app-binder/task/GetShieldedAddressTask.ts
```diff
@@ -0,0 +1,72 @@
...
+  readonly derivationPath: string;
+  readonly checkOnDevice?: boolean;
+};
+
+export type GetShieldedAddressTaskResult = DmkResult<
+  { readonly address: string },
+  CommandErrorResult<ZcashErrorCodes>["error"]
+>;
+
+function deriveShieldedAddressPaths(
```
cted-ledger, line 24: `GetFullViewingKeyTask.deriveUfvkAccountPaths` already performs this same reduction (5-level transparent path → 3-level account path) two files over. Two independent copies of one host-side derivation rule will drift the day the rule changes.

`how-to-create-signer.md` §Best Practices asks to follow established patterns; a shared helper in the package's `utils/`, consumed by both tasks, is the natural home. Worth flagging explicitly because SonarCloud will not: it measures duplication *within* new code, so it cannot see that this new code duplicates existing code.

packages/signer/signer-zcash/src/internal/app-binder/command/GetShieldedAddressCommand.ts
```diff
@@ -0,0 +1,144 @@
...
+    const apduArgs: ApduBuilderArgs = {
+      cla: ZCASH_CLA,
+      ins: GET_SHIELDED_ADDRESS_INS,
+      p1: this.args.checkOnDevice ? 0x01 : 0x00,
+      p2: 0x00,
+    };
+
+    const builder = new ApduBuilder(apduArgs);
+
+    const orchardPath = DerivationPathUtils.splitPath(
```
cted-ledger, line 69: `DerivationPathUtils.splitPath` throws a bare `Error("invalid number provided")` when a component is not numeric, and `getApdu()` has no error channel — so the throw escapes the `CommandResult` contract and reaches the device action as an unhandled exception instead of a typed error.

The task guards the number of levels but not that each level parses, so `foo'/133'/0'/0/0` reaches this line. `GetFullViewingKeyCommand` shares the weakness, so this is not a regression introduced here — but the validation suggested on `GetShieldedAddressTask.ts` would close it for both.

packages/signer/signer-zcash/src/internal/app-binder/command/GetShieldedAddressCommand.ts
```diff
@@ -0,0 +1,144 @@
...
+    return builder.build();
+  }
+
+  parseResponse(
+    apduResponse: ApduResponse,
+  ): CommandResult<GetShieldedAddressCommandResponse, ZcashErrorCodes> {
+    return Maybe.fromNullable(
+      this.errorHelper.getError(apduResponse),
+    ).orDefaultLazy(() => {
+      if (apduResponse.data.length === APDU_MAX_PAYLOAD) {
```
cted-ledger, line 95: Treating `data.length === APDU_MAX_PAYLOAD` as the continuation signal is a heuristic: it detects only an exactly-full frame, and it would misfire on a legitimate response that happens to measure `APDU_MAX_PAYLOAD`.

Given the firmware returns a UA with a single Orchard receiver (~106 bytes), the case cannot arise today — so this is fine as shipped. What is missing is the reasoning: a one-line comment saying that a UA never spans more than one frame, so a full frame means the firmware changed, would make the assumption auditable (§Best Practices #6 — document complex logic). `GetFullViewingKeyTask` solves the general case with an explicit assembly loop if you prefer symmetry over the guard.

packages/signer/signer-zcash/src/internal/app-binder/ZcashAppBinder.ts
```diff
@@ -69,6 +71,29 @@ export class ZcashAppBinder {
       }),
     });
   }
+  getShieldedAddress(args: {
```
cted-ledger, line 75: Missing blank line between `getAddress` and this method — every other method pair in the file is separated by one.

Worth noting Prettier will not fix it on commit: it preserves blank lines between class members but never inserts them, which is why the format check is green. Needs a manual edit.

packages/signer/signer-zcash/src/api/SignerZcash.ts
```diff
@@ -24,6 +25,11 @@ export interface SignerZcash {
     options?: FullViewingKeyOptions,
   ) => GetFullViewingKeyDAReturnType;
 
+  getShieldedAddress: (
```
cted-ledger, line 36: No JSDoc on this new public API entry, where §Best Practices #6 of `how-to-create-signer.md` asks for documented public APIs.

This signature needs it more than its neighbours do: `derivationPath` is the **transparent** 5-level path, and the Orchard path is silently derived from it by the task. Nothing in `(derivationPath: string, options?: AddressOptions)` lets a caller guess that, nor that a 3-level ZIP32 path will be rejected.

The sibling methods have no JSDoc either, so this is a convention the package has not yet adopted rather than something this PR broke — but a new entry point is a good place to start.

.changeset/signer-zcash-get-shielded-address.md
vladyslavchupovskiy-ext-art: Thanks ! Done

packages/signer/signer-zcash/src/internal/app-binder/command/GetShieldedAddressCommand.test.ts
vladyslavchupovskiy-ext-art: Done dooner !

packages/signer/signer-zcash/src/internal/app-binder/command/GetShieldedAddressCommand.test.ts
vladyslavchupovskiy-ext-art: Yeah thanks for that, didn't knew. Done !

packages/signer/signer-zcash/src/internal/app-binder/task/GetShieldedAddressTask.ts
vladyslavchupovskiy-ext-art: Added host-side validation in `deriveShieldedAddressPaths`:
- purpose must be exactly `44'`
- account component must be hardened (ends with `'`)
- all five components must match `/^\d+'?$/` — this also closes the `splitPath`  for this command

```ts
const isValidBip32Component = (c: string): boolean => /^\d+'?$/.test(c);
if (!components.every(isValidBip32Component)) {
  return {
    error: new InvalidStatusWordError(
      `Shielded address derivation path contains non-numeric components (got "${derivationPath}")`,
    ),
  };
}
```
GetFullViewingKeyTask shares the same weakness but it is pre-existing. Applied the same numeric guard there as well for consistency.

Worth flagging for a follow-up: the root fix belongs in `DerivationPathUtils.splitPath` itself — it should return a `Result` type rather than throwing a bare `Error`. That would fix every caller across all signers in one place without needing per-task guards. Leaving that for a dedicated signer-utils PR since touching it here would cascade across the entire codebase.

packages/signer/signer-zcash/src/internal/app-binder/task/GetShieldedAddressTask.ts
vladyslavchupovskiy-ext-art: Thanks. Done

packages/signer/signer-zcash/src/internal/app-binder/ZcashAppBinder.ts
vladyslavchupovskiy-ext-art: Thanks Done !

packages/signer/signer-zcash/src/api/SignerZcash.ts
vladyslavchupovskiy-ext-art: Done

packages/signer/signer-zcash/src/internal/app-binder/command/GetShieldedAddressCommand.ts
vladyslavchupovskiy-ext-art: Yeah agree , left oneliner in new commit

packages/signer/signer-zcash/src/internal/app-binder/command/GetShieldedAddressCommand.ts
vladyslavchupovskiy-ext-art: Closed by the `isValidBip32Component` guard added in `deriveShieldedAddressPaths` — see reply on the `GetShieldedAddressTask.ts` thread for the full context and the follow-up note on `DerivationPathUtils.splitPath`.

packages/signer/signer-zcash/src/internal/app-binder/task/GetShieldedAddressTask.ts
```diff
@@ -0,0 +1,102 @@
...
+      ),
+    };
+  }
+  if (!coin) {
+    return {
+      error: new InvalidStatusWordError(
+        `Shielded address derivation path is malformed (got "${derivationPath}")`,
+      ),
+    };
+  }
```
cted-ledger: The validation covers the purpose and the account hardening, which closes the case I raised. Two device-side constraints stay outside the net — same class of problem, non-blocking, but you are already in this code:

**Coin type hardening.** The `!coin` check here only tests presence. The ZIP32 branch of `check_bip44_compliance` requires `path[1] == ZCASH_BIP44_COIN_TYPE | HARDENED` (`app-zcash src/utils.rs:196`), so `44'/133/0'/0/0` derives `32'/133/0'` and gets rejected by the device. A `coin.endsWith("'")` is enough and stays network-agnostic: the value itself is 133 on mainnet and 1 on testnet (`src/consts.rs:15-18`), so it must not be pinned host-side.

**Change component.** The transparent path is validated with `Bip44CheckMode::Full { is_change_path: false }`, which requires `change == 0` exactly (`src/utils.rs:231,238-241`). So `44'/133'/0'/1/0` pays the full round-trip for an opaque `6A80`.

Sonar nit (line 56, the MINOR from the analysis): `!account?.endsWith("'")` replaces `!account || !account.endsWith("'")`.

packages/signer/signer-zcash/src/internal/app-binder/task/GetFullViewingKeyTask.ts
```diff
@@ -165,8 +166,15 @@ function deriveUfvkAccountPaths(
     };
   }
   const [, coinType, account] = components;
+  if (!coinType || !account) {
+    return {
+      error: new InvalidStatusWordError(
+        `Zcash UFVK derivation path components are malformed (got "${derivationPath}")`,
+      ),
+    };
+  }
```
cted-ledger, line 175: Gap between the reply and the code: the numeric guard was not applied here, only a nullish check. `"foo'"` is truthy, so `foo'/133'/0'` still passes this function and throws inside `splitPath`, outside the `CommandResult` channel.

Pre-existing and out of scope for this PR — I am noting it so the thread does not read as if it were covered. It mostly reinforces the follow-up you propose: having `DerivationPathUtils.splitPath` return a `Result` would fix both tasks, and every other signer, without a per-caller guard.

cted-ledger (commented):
Thanks — both blockers are cleared, and I re-checked the rest against the code rather than the replies. The `u1u2h4ce…` vector does match `test_get_orchard_uaddress_no_confirm` (`tests/standalone/test_pubkey_cmd.py:246-251`, `feat/zcash-integration/ironwood`), the docs are updated on both sides, and the three sibling suites are extended — including the `VerifyAddress` case on the binder.

**The PR description did not follow**, and it is the one thing I am asking you to fix before approval. It still carries `LIVE-33571` — a merged hyperliquid ADR — instead of `LIVE-35012`; it still says the UA "bundles both receivers", which your own changeset now contradicts; it announces a bump for `@ledgerhq/signer-zcash`; and the Documentation box is still unchecked with "no public docs affected" while this commit modifies `README.md` and `zcash.mdx`.

**SonarCloud** reports 2 new code smells introduced by the commit, quality gate green nonetheless: a MAJOR on `GetShieldedAddressTask.test.ts:64` (parameterize the 6 validation tests rather than repeat them) and a MINOR picked up inline below.

Doc nit: `zcash.mdx` says "Transparent derivation path (ZIP-44)" — that is BIP-44.

packages/signer/signer-zcash/src/internal/app-binder/task/GetShieldedAddressTask.ts
vladyslavchupovskiy-ext-art: Added host-side validation in `deriveShieldedAddressPaths`:
- purpose must be exactly `44'`
- account component must be hardened (ends with `'`)
- all five components must match `/^\d+'?$/` — this also closes the `splitPath` gap from #6 for this command

```ts
const isValidBip32Component = (c: string): boolean => /^\d+'?$/.test(c);
if (!components.every(isValidBip32Component)) {
  return {
    error: new InvalidStatusWordError(
      `Shielded address derivation path contains non-numeric components (got "${derivationPath}")`,
    ),
  };
}
```
GetFullViewingKeyTask shares the same weakness but it is pre-existing. Applied the same numeric guard there as well for consistency.

Worth flagging for a follow-up: the root fix belongs in `DerivationPathUtils.splitPath` itself — it should return a `Result` type rather than throwing a bare `Error`. That would fix every caller across all signers in one place without needing per-task guards. Leaving that for a dedicated signer-utils PR since touching it here would cascade across the entire codebase.

packages/signer/signer-zcash/src/internal/app-binder/task/GetShieldedAddressTask.ts
vladyslavchupovskiy-ext-art: Fixed, thanks for the clear firmware references.

One thought: since purpose, coin type, and account hardening are required by BIP44 across all signers, would it make sense to have a shared utility (e.g. in signer-utils) that validates which path components must be hardened — something like validateHardenedComponents(components, [0, 1, 2])? Each signer's task would still declare its own rules, but the check itself would be centralised rather than repeated per-signer. Same family of problem as the splitPath Result-type follow-up really — both point toward signer-utils needing a richer validation layer

packages/signer/signer-zcash/src/internal/app-binder/task/GetFullViewingKeyTask.ts
vladyslavchupovskiy-ext-art: Yeah agree, we can also potentially combine this finding with my pitch in the first comment of this 2nd review , 
maybe we can refine a ticket for that and leave it in the backlog, so when someone would be free can jump on it and do this

cted-ledger (approved):

fAnselmi-Ledger (changes requested):
LGTM but pls squash those patch commits into the original

fAnselmi-Ledger (approved):

PR #1701, opened 2026-07-31
🐛 (manager-api) [NO-ISSUE]: Handle no available update

benruseau:
### 📝 Description

This PR fixes the handling of Manager API responses when checking for the latest available firmware update.

Previously, the absence of an available firmware update could be treated the same way as an error response from the Manager API. This made it impossible to reliably differentiate between:

- a valid response where no firmware update is available;
- an actual failure, such as an HTTP/network error or an invalid Manager API response.

The fix introduces an explicit “no available update” case so callers can safely continue when the Manager API successfully reports that no update exists, while still preserving error handling for real failures.

### ❓ Context

- **Feature**: Manager API firmware metadata retrieval

### ✅ Checklist

- [ ] ~~**Documentation is up-to-date**~~
  - Firmware metadata retrieval now distinguishes “no available update” from Manager API errors.
  - Device metadata flows can continue normally when the device is already up to date.
  - HTTP/network errors and invalid Manager API responses are still handled as failures.

### DIFF
--- a/packages/device-management-kit/src/internal/manager-api/data/HttpManagerApiDataSource.ts
+++ b/packages/device-management-kit/src/internal/manager-api/data/HttpManagerApiDataSource.ts
@@ -1,5 +1,5 @@
 import { inject, injectable } from "inversify";
-import { EitherAsync } from "purify-ts";
+import { EitherAsync, Just, Maybe, Nothing } from "purify-ts";
 
 import { type DmkConfig } from "@api/DmkConfig";
 import { DmkNetworkClient } from "@api/network/DmkNetworkClient";
@@ -169,7 +169,7 @@ export class HttpManagerApiDataSource implements ManagerApiDataSource {
 
   getLatestFirmwareVersion(
     params: GetLatestFirmwareVersionParams,
-  ): EitherAsync<HttpFetchApiError, OsuFirmware> {
+  ): EitherAsync<HttpFetchApiError, Maybe<OsuFirmware>> {
     const livecommonversion = "34.27.0"; // Legacy parameter that should just be a too old
     const { currentFinalFirmwareId, deviceId } = params;
     return EitherAsync(() =>
@@ -330,13 +330,15 @@ export class HttpManagerApiDataSource implements ManagerApiDataSource {
           if (
             typeof mcu !== "object" ||
             typeof mcu.id !== "number" ||
-            typeof mcu.name !== "string"
+            typeof mcu.name !== "string" ||
+            typeof mcu.from_bootloader_version !== "string"
           ) {
             throw new Error(`Incomplete MCU version: ${JSON.stringify(mcu)}`);
           }
           const ret: McuFirmware = {
             id: mcu.id,
             name: mcu.name,
+            fromBootloaderVersion: mcu.from_bootloader_version,
           };
           return ret;
         }),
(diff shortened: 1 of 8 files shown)


packages/device-management-kit/src/internal/manager-api/data/HttpManagerApiDataSource.ts
```diff
@@ -402,8 +404,12 @@ export class HttpManagerApiDataSource implements ManagerApiDataSource {
 
   private mapLatestFirmwareDto(
     latestFirmware: LatestFirmwareOsuVersionResponseDto,
-  ): EitherAsync<Error, OsuFirmware> {
-    return EitherAsync<Error, FirmwareOsuVersionDto>(() => {
+  ): EitherAsync<Error, Maybe<OsuFirmware>> {
+    return EitherAsync<Error, FirmwareOsuVersionDto | null>(() => {
+      if (latestFirmware.result === "null") {
```
benruseau, line 413: This is how we know that there is no update available (see in [ledger-live](https://github.com/LedgerHQ/ledger-live/blob/develop/libs/device-core/src/managerApi/repositories/HttpManagerApiRepository.ts#L59-L61))

packages/device-management-kit/src/internal/manager-api/model/Firmware.ts
```diff
@@ -1,5 +1,6 @@
 export type OsuFirmware = {
   id: number;
+  notes: string | null;
```
benruseau, line 3: We will need this for the OS updates (used to show a description of the OS update in LW)

packages/device-management-kit/src/internal/manager-api/model/Firmware.ts
```diff
@@ -21,4 +22,5 @@ export type FinalFirmware = {
 export type McuFirmware = {
   id: number;
   name: string;
+  fromBootloaderVersion: string;
```
benruseau, line 25: We will need this to flash the booloader

packages/device-management-kit/src/internal/manager-api/service/ManagerApiService.ts
```diff
@@ -79,7 +79,7 @@ export interface ManagerApiService {
   getLatestFirmwareVersion(
     currentFirmware: FinalFirmware,
     deviceVersion: DeviceVersion,
-  ): EitherAsync<HttpFetchApiError, OsuFirmware>;
+  ): EitherAsync<HttpFetchApiError, Maybe<OsuFirmware>>;
```
benruseau, line 82: New signature: maybe return an `OsuFirmware` if an OS update is available, or `Nothing`

packages/device-management-kit/src/internal/manager-api/data/HttpManagerApiDataSource.ts
mbertin-ledger: [COULD] Maybe add a doc comment (if you have to mention it, let it in the code ;) )

packages/device-management-kit/src/internal/manager-api/data/HttpManagerApiDataSource.ts
```diff
@@ -402,8 +404,12 @@ export class HttpManagerApiDataSource implements ManagerApiDataSource {
...
+        // Used to differentiate between no update available and an error.
+        return Promise.resolve(null);
```
mbertin-ledger, line 414: [ASK] Null means no update, undefined is error ?

mbertin-ledger (approved):

packages/device-management-kit/src/internal/manager-api/data/HttpManagerApiDataSource.ts
benruseau: Yes!

PR #1704, opened 2026-08-03
🐛 (signer-solana) [DSDK-1406]: Forward cosigners for txc

fAnselmi-Ledger:
### 📝 Description

**Problem**

Any Solana transaction requiring more than one signature reaches Transaction Check with every signature slot zeroed. The pending signer's slot is tolerated, but every other empty slot is flagged as an invalid signature, producing Critical Threat Detected verdicts on legitimate multi-signer transactions like LP position creation on Meteora or Orca.

Root cause: signTransaction only receives message bytes (the Solana app signs a message, not a full transaction), so `SolanaTransactionCheckLoader` reconstructs the wire transaction with all signature slots empty. Co-signer signatures that are already present on the caller's `VersionedTransaction` object are discarded at the API boundary with no way to pass them through.

**Change**

Adds `serializedTransactionForCheck?: Uint8Array` to `SolanaTransactionOptionalConfig`. When provided, the loader recovers co-signer signatures from the blob and writes them at their positional offsets in the Transaction Check payload instead of zero-filling.

The field never reaches the device. transaction remains the sole source of truth for signing, clear-signing, and fingerprinting.

Plumbing (signer-solana)

- New optional field on `SolanaTransactionOptionalConfig` 
- Threaded through `SignTransactionDeviceAction` and `ProvideTransactionCheckTask` into `getContexts`

Recovery logic (context-module)

- `SolanaTransactionCheckRequest` and the input codec extended 
- `wrapMessageAsTransaction` writes recovered signatures at their positional offsets
- Validation before accepting anything: signature count matches message header, message region length matches, message region bytes match 
- Any parse or validation failure degrades to zero-fill and never throws

### ❓ Context

- **JIRA or GitHub link**: [DSDK-1406](https://ledgerhq.atlassian.net/browse/DSDK-1406)

### ✅ Checklist

  - _list of the changes_

---

### DIFF
--- a/.changeset/modern-lemons-dig.md
+++ b/.changeset/modern-lemons-dig.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/device-signer-kit-solana": patch
+---
+
+Forward cosigners for Solana txc
--- a/.changeset/slow-rabbits-enjoy.md
+++ b/.changeset/slow-rabbits-enjoy.md
@@ -1,0 +1,5 @@
+---
+"@ledgerhq/context-module": patch
+---
+
+Forward cosigners for Solana txc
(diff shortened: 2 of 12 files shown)


smartine-ledger (commented):
One design question:  Why not accept the full serialized transaction directly as the transaction parameter and extract the message bytes internally? That way the caller wouldn't need to worry about the distinction, and co-signer signatures would be available automatically without an extra field.

fAnselmi-Ledger:
closing this, a fix has been made downstream

PR #1739, opened 2026-08-14
✨ (contacts-kit) [DSDK-1377]: Register an external address

daniel-choinski-ledger:
### 📝 Description

Implements **DSDK-1377: Register an external address**, the first operation of the new `@ledgerhq/device-contacts-kit` (Address Book). It adds the full vertical slice for the firmware **REGISTER IDENTITY** command plus the shared Contacts protocol foundation that the remaining Address Book operations will reuse.

**Public API**

```ts
const contactsManager = new ContactsManagerBuilder({ dmk, sessionId, appName: "Ethereum" }).build();

const { observable } = contactsManager.registerExternalAddress({
  contactName: "Alice",
  scope: "Eth main",
  identifier: addressBytes,      // 20-byte Ethereum address
  blockchainFamily: "ethereum",
  chainId: 1n,                   // mandatory for Ethereum
  // existingContactGroup: { groupHandle, hmacProof }, // to add an address to an existing contact
  // skipOpenApp: true,          // skip only the open-app step; the version guard still runs
});
// on completion → { mode, …echoed input, groupHandle, hmacProof, hmacRest }
```

### ❓ Context

- **JIRA or GitHub link**: [DSDK-1377]

- **Feature**:

### ✅ Checklist

- [X] **Covered by automatic tests**
- [X] **Changeset is provided** 
- [X] **Documentation is up-to-date** 
  - New public API ContactsManager.registerExternalAddress() and its DA/model types on @ledgerhq/device-contacts-kit (previously an empty scaffold)
  - New shared Contacts protocol foundation (TLV, framing, errors, validation) under internal.
  - New apps/sample Contacts page (provider, view, route, sidebar entry) that persists proofs to localStorage (dmk-sample-contacts).
 
---

[DSDK-1377]: https://ledgerhq.atlassian.net/browse/DSDK-1377?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/packages/device-contacts-kit/src/internal/app-binder/device-action/RegisterExternalAddress/RegisterExternalAddressDeviceAction.ts
+++ b/packages/device-contacts-kit/src/internal/app-binder/device-action/RegisterExternalAddress/RegisterExternalAddressDeviceAction.ts
@@ -1,0 +1,326 @@
+import {
+  type DeviceActionStateMachine,
+  type InternalApi,
+  isSuccessCommandResult,
+  OpenAppDeviceAction,
+  type StateMachineTypes,
+  UnknownDAError,
+  UserInteractionRequired,
+  WaitForAppAndVersionDeviceAction,
+  XStateDeviceAction,
+} from "@ledgerhq/device-management-kit";
+import { Left, Right } from "purify-ts";
+import { assign, fromPromise, setup } from "xstate";
+
+import {
+  type RegisterExternalAddressDAError,
+  type RegisterExternalAddressDAInput,
+  type RegisterExternalAddressDAIntermediateValue,
+  type RegisterExternalAddressDAInternalState,
+  type RegisterExternalAddressDAOutput,
+} from "@api/app-binder/RegisterExternalAddressDeviceActionTypes";
+import {
+  isContactsSupportedForSession,
+  type RunningApp,
+} from "@internal/app-binder/isContactsSupportedForSession";
+import { ContactsVersionRequirementError } from "@internal/app-binder/model/contactsErrors";
+import {
+  type RegisterIdentityProofs,
+  SendRegisterIdentityTask,
+} from "@internal/app-binder/task/SendRegisterIdentityTask";
+
+import { validateRegisterExternalAddressInput } from "./validateRegisterExternalAddressInput";
+
+export type RegisterExternalAddressMachineDependencies = {
+  readonly isSupported: (app: RunningApp) => boolean;
+  readonly registerIdentity: (
+    input: RegisterExternalAddressDAInput,
+  ) => Promise<Awaited<ReturnType<SendRegisterIdentityTask["run"]>>>;
+};
+
+export class RegisterExternalAddressDeviceAction extends XStateDeviceAction<
+  RegisterExternalAddressDAOutput,
+  RegisterExternalAddressDAInput,
+  RegisterExternalAddressDAError,
+  RegisterExternalAddressDAIntermediateValue,
+  RegisterExternalAddressDAInternalState
+> {
+  makeStateMachine(
+    internalApi: InternalApi,
+  ): DeviceActionStateMachine<
+    RegisterExternalAddressDAOutput,
+    RegisterExternalAddressDAInput,
+    RegisterExternalAddressDAError,
+    RegisterExternalAddressDAIntermediateValue,
+    RegisterExternalAddressDAInternalState
+  > {
+    type types = StateMachineTypes<
+      RegisterExternalAddressDAOutput,
+      RegisterExternalAddressDAInput,
+      RegisterExternalAddressDAError,
+      RegisterExternalAddressDAIntermediateValue,
+      RegisterExternalAddressDAInternalState
+    >;
+
+    const { isSupported, registerIdentity } =
+      this.extractDependencies(internalApi);
+    const appName = this.input.appName;
+
+    return setup({
+      types: {
+        input: {} as types["input"],
+        context: {} as types["context"],
+        output: {} as types["output"],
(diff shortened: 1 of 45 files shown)


packages/device-contacts-kit/src/internal/use-cases/RegisterExternalAddressUseCase.ts
OlivierFreyssinet, line 1: [SHOULD] Not throw, move validation in device action.

It might be too implicit right now, but public APIs in this repo generally do not throw.

`ContactsManager.registerExternalAddress()` must not throw for invalid caller input.

The public API returns a `DeviceAction`, but `RegisterExternalAddressUseCase.execute()` performs synchronous validation and throws `ContactsValidationError`, forcing consumers to use try/catch rather than handling a typed terminal DA error.

Please move all validation into the device action (once the app is opened) and include `ContactsValidationError` in `RegisterExternalAddressDAError`, so every failure reaches the observable as a specific error state.

packages/device-contacts-kit/src/internal/app-binder/services/contactsTlvSerializer.ts
```diff
@@ -0,0 +1,125 @@
...
+
+function writeTag(builder: ByteArrayBuilder, tag: number): void {
+  if (tag < 0x80) {
+    builder.add8BitUIntToData(tag);
+  } else if (tag <= 0xff) {
+    builder.add16BitUIntToData(LONG_TAG_PREFIX_HIGH_BYTE + tag);
+  } else {
+    throw new Error(`TLV tag ${tag} exceeds supported single-byte range`);
+  }
+}
```
OlivierFreyssinet, line 51: [SHOULD]
Tags are always encoded as their raw one-byte value per the [Address Book specification](https://ledgerhq.atlassian.net/wiki/spaces/FW/pages/6992035925/Address+Book+Final+Specifications#TLV-encoding), so there is no need to branch on or transform the tag value.

packages/device-contacts-kit/src/internal/app-binder/device-action/RegisterExternalAddress/RegisterExternalAddressDeviceAction.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  extractDependencies(
+    internalApi: InternalApi,
+  ): RegisterExternalAddressMachineDependencies {
+    const isSupported = () => {
+      const deviceState = internalApi.getDeviceSessionState();
+      const version =
+        "currentApp" in deviceState
+          ? deviceState.currentApp?.version
+          : undefined;
```
OlivierFreyssinet: [SHOULD] Not use this data which can be stale (for instance if the session refresher of the DMK isn't enabled and the app is changed outside of DMK, and if open app is skipped)
Run `WaitForAppAndVersionDeviceAction` before the guard and evaluate support from its fresh { name, version } result.

daniel-choinski-ledger: Great insight, thanks!

packages/device-contacts-kit/src/internal/app-binder/device-action/RegisterExternalAddress/RegisterExternalAddressDeviceAction.ts
```diff
@@ -0,0 +1,251 @@
...
+      if (version === undefined) return false;
+      return isContactsSupportedForSession(internalApi, { version });
```
OlivierFreyssinet: I'm not sure that this should be used actually, as in many scenarios, session data does not contain the OS version (OS version cannot be obtained when an app is opened, so by default if the session starts with an open app this info is unknown).
I don't know yet how to solve this, but if we really want to strictly enforce the min OS version in here we probably have to first go to the dashboard, get the os version, then open the app ...

OlivierFreyssinet: So maybe we should actually do this first in this device action:
- check if device session state is populated with an OS version
  - if it is, continue (verify that the OS is supported)
  - else, quit app, get os version, save to session state, reopen app

This logic, which will be required by all Contacts Device Actions could be extracted into a standalone reusable device action.

packages/device-contacts-kit/src/internal/app-binder/device-action/RegisterExternalAddress/RegisterExternalAddressDeviceAction.ts
```diff
@@ -0,0 +1,316 @@
...
+            input: () => ({}),
+            onDone: {
+              target: "CheckAppAndVersion",
+              actions: assign({
+                _internalState: ({ event, context }) =>
+                  event.output.caseOf<RegisterExternalAddressDAInternalState>({
+                    Right: (appAndVersion) => ({
+                      ...context._internalState,
+                      appAndVersion: {
+                        name: appAndVersion.name,
```
OlivierFreyssinet, line 215: [SHOULD] `WaitForAppAndVersion` might emit an intermediate value, you have to propagate it here.
For instance if the device is locked, there will be a "unlock" user required interaction, your device action should propagate it so the client consuming the DA is able to display a "please unlock your device" UI.
You'll find many example of this in the codebase (a DA invoking another DA and propagating the intermediate value).

OlivierFreyssinet (approved):

PR #1790, opened 2026-08-25
✨ (signer-xrp) [DSDK-1441]: Add SignTransaction use case and sample integration

aussedatlo:
> [!NOTE]
> Top of a three-PR stack, on top of #1789 ([DSDK-1440](https://ledgerhq.atlassian.net/browse/DSDK-1440)) — this PR targets `feat/dsdk-1440-xrp-send-sign-transaction-task`, so review only the last two commits. Retarget to `develop` as the stack merges down.

### 📝 Description

Wires `SendSignTransactionTask` through the full stack so a transaction can be signed from the sample app.

`XrpAppBinder.signTransaction` now builds its device action through the exported `SignTransactionDeviceActionFactory`, which runs the task. Transactions are therefore chunked and carry their derivation path — **the scaffold task this replaces did neither**, so signing could only ever have worked by accident for payloads under one APDU. Adds a `DmkLoggerFactory` binding, which is what the task needs for its logger, and defaults `skipOpenApp` to `false` in the use case.

```typescript
const { observable, cancel } = signerXrp.signTransaction(
  "44'/144'/0'/0/0",
  transactionBytes,
);
```

The sample now parses its hex blob with `hexaStringToBuffer` instead of hand-rolled slicing, rejects malformed input, and defaults to XRP's coin type rather than Bitcoin's.

#### ⚠️ Two deviations from the ticket

**1. No `curve` option.** The ticket scopes `curve?` in `TransactionOptions` and a curve selector in the sample. ed25519 is not exposed for the reason given in #1788: `sign_transaction.c` hands the chosen curve to `bip32_derive_init_privkey_256`, so it cannot succeed on the device.

**2. Rejecting surfaces `6985`, not `6982`.** The ticket says `6982`; `reject_transaction()` in `app-xrp` sends `0x6985`, which the e2e test pins.

### ❓ Context

- **JIRA or GitHub link**: [DSDK-1441](https://ledgerhq.atlassian.net/browse/DSDK-1441)
- **Feature**: XRP signer — sign transaction use case and sample integration

### ✅ Checklist

  - Unit: `SignTransactionUseCase.test.ts` (defaults, forwarding) and updated `XrpAppBinder` cases.
  - E2E: new `apps/sample/playwright/cases/signers/xrp/sign-transaction.spec.ts` — three cases against the **real** XRP app on Speculos (stax `1.10.1` / XRP `2.6.1`): a payment fitting one APDU, a payment carrying a memo, and a rejection. Both blobs are `ripple-binary-codec`-encoded Payments pinned as fixtures, so the sample needs no new dependency. The memo one is 425 bytes — 446 once the derivation path is prepended, so **two chunks with the first one full**, which is what exercises DSDK-1440's loop against the device.
  - All 11 XRP e2e specs and 76 unit tests pass locally.
- [x] **Changeset is provided** — `.changeset/xrp-sign-transaction-use-case.md`
  - `SignerXrp.signTransaction` now actually works for payloads over one APDU.
  - `XrpAppBinder` gains a constructor parameter (the logger factory), so any direct instantiation needs updating.
  - The scaffold `SignTransactionTask` is deleted.
  - Sample app: corrected derivation path default and stricter hex parsing.
  - Worth knowing for whoever writes further device tests: rejecting on Stax needs **two** taps — "Reject", then "Yes, reject" on the modal it opens. Without the second the APDU goes unanswered and the mock server's proxy times out into a `6d00`, which an assertion on the error type alone would accept as a rejection.

[DSDK-1440]: https://ledgerhq.atlassian.net/browse/DSDK-1440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[DSDK-1441]: https://ledgerhq.atlassian.net/browse/DSDK-1441?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/signer/signer-xrp/src/internal/app-binder/device-action/SignTransactionDeviceActionFactory.ts
```diff
@@ -0,0 +1,49 @@
...
+      task: async (internalApi) =>
+        new SendSignTransactionTask(internalApi, {
+          derivationPath: args.derivationPath,
+          serializedTransaction: args.transaction,
+          loggerFactory: args.loggerFactory,
+        }).run(),
+      appName: APP_NAME,
+      requiredUserInteraction: UserInteractionRequired.SignTransaction,
+      skipOpenApp: args.skipOpenApp,
+    },
```
fAnselmi-Ledger, line 48: [should] missing logger here

fAnselmi-Ledger (approved):

PR #1802, opened 2026-08-27
✨ (sample) [NO-ISSUE]: Dock a live device screen in the sidebar

aussedatlo:
### 📝 Description

https://github.com/user-attachments/assets/4172d47a-2764-4822-8efa-185219f18fe3

The device screen used to be a draggable pastille floating over the page, VNC-only, so it worked with the Speculos transport and not at all with the mock server. It also sat on top of the content by default and clamped the wrong size, so it could hang off screen.

Now it's a sticky panel in the sidebar, collapsible, shown while a device is connected. It's a flex sibling instead of an overlay, so no drag handling and no bounds maths.

Both transports use the same thing: polled screenshots + `button`/`finger` calls. Speculos serves those endpoints directly and the mock server proxies to them, so the only difference is the URL — that's what `ScreenApi` is. VNC, `react-vnc` and the VNC URL setting are removed.

Details worth a look:

- press/release are separate calls on pointer down/up, so a hold is an actual hold (Stax/Flex need one to confirm). Pointer capture + release on unmount so we can't leave the device pressed.
- no per-model size table: the PNG's own dimensions size the frame and map taps to device pixels. Only the input mode (touch vs buttons) is looked up.
- no app running → no Speculos instance → proxy returns 409, and we show the device's record instead. No need to check `currentApp`.
- if a frame can't be decoded we say so instead of rendering a broken image.

Also in here, both needed to make the above work:

- `device-mock-server`: the Speculos passthrough read every response with `text()`, which destroyed PNGs, so `/screenshot` could never work. Reads bytes now. The old tests passed only because they stubbed the body as `"PNG-BYTES"`.
- `mockserver-client`: adds `getScreenshot` / `pressButton` / `touchScreen` next to `getSpeculos`.

Last commit seeds a Stax, a Flex and a Nano X into new mock sessions with all the signer apps, so you don't have to hand-build a device and guess a firmware/app pair that Speculinho can resolve.

### ❓ Context

- **Feature**: live device screen in the sample for both transports, replacing the VNC viewer.

### ✅ Checklist

- [x] **Covered by automatic tests** — unit tests for the binary relay (incl. a non-UTF-8 round-trip) and the new `MockClient` methods. The panel is sample UI, not covered by Playwright; tested by hand against real Speculos pods.
- [x] **Changeset is provided** — minor for `@ledgerhq/device-mockserver-client`. Apps are private, no changeset needed.
- [x] **Documentation is up-to-date** — nothing user-facing to update.
  - `apps/sample`: VNC pastille + setting gone, device screen is a sidebar panel, new mock sessions come with 3 devices.
  - `apps/device-mock-server`: Speculos passthrough relays bytes. **Needs a release** — until the hosted instance is redeployed `/screenshot` still returns mangled PNGs and the panel shows the decode message.
  - `packages/device-mockserver-client`: 3 new methods, nothing breaking.
  - `selectOrderedConnectedDevices` is memoized now, kills a react-redux warning from its 4 call sites.

Tested on real pods: Stax 400×672 + hold works, Nano X 128×64 + Left/Both/Right works, both fall back to the device card when the app closes, all seeded app versions open with 9000.


OlivierFreyssinet (approved):
🔥

PR #1901, opened 2026-09-18
♻️ (os-update) [NO-ISSUE]: Flash mcu device action

benruseau:
### 📝 Description

`FlashMcuDeviceAction` could only be used as a step of an OS update: it required a
`finalFirmware` to pick the MCU from. This adds a second mode so a device stuck in
bootloader mode, with no known update path, can still be recovered.

`FlashMcuDAInput` becomes a discriminated union:

- `{ mode: "osUpdate", finalFirmware }` — previous behaviour.
- `{ mode: "bootloaderRecovery" }` — resolves the final firmware from what the device
  reports (`seVersion` / `seTargetId` against the Manager API), and falls back to the
  bootloader version alone when the device reports no secure element firmware.

Two bugfixes come with it:

- The forced MCU hop table is now per mode. An OS update only aliases a bootloader
  reporting no version; previously a device reporting bootloader `0.6`, `0.7` or `0.9`
  skipped the MCU catalog entirely and ignored the firmware being installed.
- Bootloader versions are now compared coerced (`1.16` vs `1.16.0`), since the MCU
  catalog and the devices do not agree on the number of segments. The strict string
  equality could send a device on a pointless bootloader hop.

### ❓ Context

- **Feature**: MCU flashing / bootloader recovery

### ✅ Checklist

  - **Breaking:** `FlashMcuDeviceAction` (exported from `@ledgerhq/dmk-ledger-wallet`)
    now requires a `mode` in its input.
  - QA: MCU flashing during an OS update should be unchanged on recent devices, but is
    worth re-checking on old bootloaders (`0.6` / `0.7` / `0.9`), where the resolved MCU
    version now comes from the catalog.
  - QA: new path — flash the MCU on a device left in bootloader mode, both with and
    without a readable secure element firmware.


aussedatlo (approved):
No e2e tests ? 😛

daniel-choinski-ledger (approved):

